PR #24, opened 2020-01-20
Changing geometry to object-orientated interface [do not merge]

rbooth200:
Here is a suggested implementation of the new interface for the geometry fit. Let me know what you think. The interface is now more consistent with the fitters interface. To see how it all works, look at the new tests. 

It still needs docs and cleaning up - e.g. I will remove / hide the fit_geometry_gaussian function.

### DIFF
--- a/frankenstein/geometry.py
+++ b/frankenstein/geometry.py
@@ -9,7 +9,7 @@ from scipy.optimize import least_squares
 from frankenstein.constants import rad_to_arcsec
 
 __all__ = ["apply_phase_shift", "deproject", "fit_geometry_gaussian",
-           "SourceGeometry"]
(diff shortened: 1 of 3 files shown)


mtazzari:
This is fine for me to rebase and merge, but you still have [do not merge] in the title: are you waiting to do something else in this PR?

rbooth200:
If you're happy great. I need to check the docs etc, will do that now. I'll merge straight after unless you want to see it again

mtazzari:
Oh right, the documentation.
For the time being, it's important that you just write complete docstrings
of classes and functions, will care about how to use the geometry objects
in tutorial/examples/cookbook later.
Please go ahead and merge when ready.

rbooth200:
I've done the doc-strings - that's all I had in mind anyway.

PR #28, opened 2020-01-22
Add unique versioning system (wip)

mtazzari:
I add a single-source of version number for robust reference.
The version number is needed for PyPI later and for reference in the docs.


mtazzari:
@rbooth200 please rebase and merge if you are OK with this.

PR #33, opened 2020-01-30
Change internal unit of rmax to arcsec

jeffjennings:
Since a fit can be performed with, e.g., `FrankFitter` without first having to convert inc and PA from the user-provided unit [deg], the same should be true for `Rmax`. The user provides it in [arcsec], while it's handled internally in [rad]. But currently this conversion has to be done externally.

So in `hankel.py`, I've kept `self._Rnk` and `self._Rmax` in [rad], and just converted `Rmax` (not `self._Rmax`) to from [deg] to [rad] in `__init__`. This allows `Rmax` to be passed into `DiscreteHankelTransform` from `FourierBesselFitter` in the user-provided [arcsec], while still effectively handling `Rmax` internally in [rad] (consistent with `self._Rnk_`). 

In `tests.py` I've confirmed this change hasn't affected the fit by successfully running `test_frank_fitter()`. I've updated all functions in `test.py` to provide `Rmax` in [arcsec].

### DIFF
--- a/frankenstein/hankel.py
+++ b/frankenstein/hankel.py
@@ -23,6 +23,8 @@ discrete varient, the Discrete Hankel Transform (DHT).
 import numpy as np
 from scipy.special import j0, j1, jn_zeros, jv
 
+from frankenstein.constants import rad_to_arcsec
+
 __all__ = ["DiscreteHankelTransform"]
 
 
@@ -57,6 +59,8 @@ class DiscreteHankelTransform(object):
         of 2*pi.
     """
     def __init__(self, Rmax, N, nu=0):
+
+        Rmax /= rad_to_arcsec 
 
         # Select the fast Bessel functions, if available.
         if nu == 0:
--- a/frankenstein/hankel.py
+++ b/frankenstein/hankel.py
@@ -60,8 +60,6 @@ class DiscreteHankelTransform(object):
     """
     def __init__(self, Rmax, N, nu=0):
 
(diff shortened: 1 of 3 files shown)


rbooth200:
Jeff, Can you make the change in FrankFitter/FourierBesselFitter. The problem is that the units of DHT need to be self-consistent. Since the uv-points are in lambda (1/radian) Rmax must be in Radians.

rbooth200:
I would actually just change the interface of the code and not the internal workings. I.e. modify the class attributes to return Rmax / arcsec and change what's passed into FrankFitter / FourierBesselFitter

jeffjennings:
I actually want to avoid changing what's passed into FrankFitter / FourierBesselFitter, because this has to be done external to those, so there has to be a "convert_units" function called before calling the code, which isn't necessary. 

I haven't changed the internal units of DHT or the units it returns (`self._Rmax` is still in [rad]); I've just changed the local `Rmax` within DHT in the same way you update 
```
    dRA *= 2. * np.pi / rad_to_arcsec
    dDec *= 2. * np.pi / rad_to_arcsec
```
in `geometry.py`. 

It also looks like, in `geometry.py`, inc and PA are taken in as rad, instead of the user-provided deg. 

I'll stop by today so we can quickly sort this.

rbooth200:
I think you miss-understood what I said

jeffjennings:
It's possible ha, easier to go through in-person this afternoon probably.

PR #54, opened 2020-02-20
Changed how geometry is specified

rbooth200:
I have changed how the geometry fit is specified in fit.py. The motivation for this change is to make it impossible to accidentality specify multiple options, which was possible before. Now the user will be 100% certain what kind of fit they will get.


rbooth200:
There are also some changes added by autopep-8

PR #55, opened 2020-02-20
Convergence diagnostics

### LINKED ISSUE #45: add plotting functions to API
Per MT's suggestion

### LINKED ISSUE #46: remove all "sol.<blah>" from make_figs
Don't pass sol to make_figs, just pass the dict entries in sol that are needed to plot.

### LINKED ISSUE #47: review all remaining inline "# TODO"s
In .py files and docs

### LINKED ISSUE #49: Add diagnostic plots for convergence
Following discussion on Wed afternoon, add two diagnostic plots to the automatically generated figures.
Purpose of these figures is to inform the user on the convergence status of the fit.

### LINKED ISSUE #50: Documentation for diagnostic figures of convergence
After implementing the figures in #49, add a short description/comment on how to interpret these figures to the documentation.

jeffjennings:
Add plotting for convergence diagnostics, other various minor changes

Fix #45 Fix #46 Fix #47 Fix #49 Fix #50

@jeffjennings: the "Fix #N" statements I added yesterday in this description were not reminders to be removed.

A "Fix #N" in a PR description or in a commit message will close the #N issue when the PR is merged. It is a bad practice to close the issues manually, and it is actually better not to unless they become irrelevant, and won't imply any change to the repo. 
Using the "Fix #N" statements ensures that the issues are closed really when they are closed (i.e. when the PR is merged). Until this PR is merged the issues are not closed. If we were not to merge this PR now, you would have closed issues that are not actually closed, with the obvious confusion. 
Secondly, by using "Fix #N" statements Github automatically adds a reference with a link to the PR or the commit where the issue was closed, in the issue page, which helps a lot debugging later on. While now you added manually "see file..." without the PR reference number, which makes it very hard to track the commit down


rbooth200:
This PR is ready to go from my point of view.

rbooth200:
Marco tells me @jeffjennings is still working on this, I'll wait to close it.

rbooth200:
I've fixed some issues that pyling was giving, but the diagnostic plots aren't working yet, apparently.

rbooth200:
Quick Q: why did you remove the pylint directives?

jeffjennings:
My mistake, I thought it had automatically added those and that they could be removed

jeffjennings:
Everything works for me on this branch now, including the diagnostic plot.

rbooth200:
I'd also suggest replacing start_iter and stop_iter with a single num_iter parameter and just print the so many iterations. Isn't that what will be wanted anyway? I think there is too much chance of error with the current parameters. 

Obviously if you ask for 1000 and it converged in 20 then we just show all of them.

rbooth200:
The pylint directives are to stop it reporting false positives

jeffjennings:
yes good point, I just pushed with update

rbooth200:
That's not quite what I had in mind. Ill

jeffjennings:
I just updated again

rbooth200:
Sure, but i was thinking of changing how those parameters work

jeffjennings:
Open to another way, but I think the current `iter_plot_range` is good; an error isn't raised if it's specified inconsistently with `max_iter`, it's just handled by frank and printed to terminal.

rbooth200:
Ok, I'll just see if I can make it less error prone.

jeffjennings:
On TODOs, did you add the TODO in `fit.py`, l.35,

`frank_path = os.path.dirname(frank.__file__)  # TODO`

and if so is the TODO still needed? 

I also don't understand `fit.py`, l.96,
```
        uv_path = model['input_output']['uvtable_filename']
        if uv_path: 
```
becaue `uv_path` will always be true. should `if uv_path` be unindented?

rbooth200:
I'll look at the Todo when I get in.

The uv_path check is to protect against empty uv table strings. Maybe dirname works properly in that case, I don't know. 

I understand that an empty UV path is an error, but you don't want the code to break at this location if the path is empty, so I just chose a sensible option.

rbooth200:
You can remove the if statement if you think that it is unnecessary, but it produces the correct behaviour as is

jeffjennings:
`frank_path = os.path.dirname(frank.__file__) # TODO` is the last TODO currently in the code, so feel free to close #47 when resolved.

jeffjennings:
> Ok, I'll just see if I can make it less error prone.

I think I see what you mean, I've rewritten and will push in a minute after testing.

jeffjennings:
just fyi your autopep8 isn't using 80 characters as the maximum line length, so when it increases the indents in some blocks, it can make lines fairly longer when they have spaces between variables where a new line could start.

rbooth200:
Can you look at my last commit. I'm now using `None` to specify a default behaviour of the full range. Otherwise its up to the user to specify a valid range

jeffjennings:
Looking at it now; please don't run autopep8 on the .json files; it screws up the readability.

jeffjennings:
It also makes the indenting on the brackets incorrect.

rbooth200:
> Looking at it now; please don't run autopep8 on the .json files; it screws up the readability.

It not autopep8, but anyway, I've turned off the auto formatting.

jeffjennings:
This makes less sense than what was there before. 

- Don't call the diag plot just to plot 0 iterations. 

- If the user can set `max_iter`in the .json, they can specify `iter_plot_range` ... just set the upper bound to large (say 2000) by default, and anyway I print the range being plotted when `make_diag_fig` is called.

rbooth200:
>This makes less sense than what was there before.
>
> -   Don't call the diag plot just to plot 0 iterations.
>
>   -  If the user can set max_iterin the .json, they can specify iter_plot_range ... just set the upper bound to large (say 2000) by default, and anyway I print the range being plotted when make_diag_fig is called.

Actually, I think its sensible to call the plot with no iterations if that is consistent with what was requrested. The figure will be set up with correct axis-labels etc, so the user will at least see what should have been there.

rbooth200:
I also think its better for the code to complete, doing something sensible, but without the diagnostic plot if a bad iteration range was chosen. This way the fit will be stored, along with the diagnostics so the user can get at them without running the code again, if they want.

rbooth200:
> just fyi your autopep8 isn't using 80 characters as the maximum line length, so when it increases the indents in some blocks, it can make lines fairly longer when they have spaces between variables where a new line could start.

I'll fix that...

jeffjennings:
> > This makes less sense than what was there before.
> > 
> > * Don't call the diag plot just to plot 0 iterations.
> > * If the user can set max_iterin the .json, they can specify iter_plot_range ... just set the upper bound to large (say 2000) by default, and anyway I print the range being plotted when make_diag_fig is called.
> 
> Actually, I think its sensible to call the plot with no iterations if that is consistent with what was requrested. The figure will be set up with correct axis-labels etc, so the user will at least see what should have been there.

The problem is that alternatively it could just look like the code is broken and doesn't populate the plot that it's supposed to, and the user is confused why (and actually, if the plotting is broken in some other way, that's then degenerate with this). I don't think it makes sense to plot an empty figure; just explicitly state in the logging message printed to the terminal that the plot won't be made. You don't have to raise an error to stop the code, just print the log message and skip the diag plot.

rbooth200:
Isn't that what the logging message is for?

rbooth200:
You don't want to skip the full diagnostic plot because the ranges are different for different sub panels

jeffjennings:
> You don't want to skip the full diagnostic plot because the ranges are different for different sub panels

That's incorrect, all plots in the diagnostic fig are produced using `iter_plot_range`

If you make that parameter `null` by default, it's also not immediately clear it needs to be a list. I'm setting the default as 
```
"iter_plot_range"  : [0,5000]
```
so that for 99% of cases it won't matter.

rbooth200:
Actually, you only use the last 100 iterations for 2 of the panels

jeffjennings:
Hey @mtazzari, @rbooth200  and I have both been contributing to this branch that closes the outstanding issues in the Code freeze milestone. Can you review and merge into master? Thanks!

mtazzari:
Ok, starting to review

rbooth200:
I've just merged in some changes needed to get tests of the full pipeline working

frank/geometry.py
```diff
@@ -22,12 +22,10 @@
    NOTE: The sign convention used here is xx.
 """
 
-import numpy as np
```
mtazzari: In general, I see that it was adopted a convention to import `frank` modules first and external modules (e.g., `numpy`, `scipy`, ...) later.
Actually, the [correct PEP8 practice](https://www.python.org/dev/peps/pep-0008/#imports) is the other way around: 
```
    1. Standard library imports.
    2. Related third party imports.
    3. Local application/library specific imports.
```
One has to ensure frank is compliant with the distribution, not the other way around.
Hopefully `np`, or `least_squares` are not objects in `frank`, but if it was the case, `import numpy as np` as it is right now would shadow our `np` object.
So, nothing to change for now as long as it works as intended.
In the future, we might just want to adopt the right convention in the whole codebase.

frank/radial_fitters.py
```diff
@@ -20,16 +20,14 @@
    of deprojected visibities.
 """
 
+from frank.hankel import DiscreteHankelTransform
```
mtazzari: here another case of importing frank before standard packages. risky

docs/install.rst
```diff
@@ -24,11 +24,13 @@ Clone the source repository from `GitHub <https://github.com/discsim/frank>`_,
 
     pip install git+https://github.com/discsim/frank.git
 
+(this is the same as `git clone https://github.com/discsim/frank.git; cd frank`).
```
mtazzari: This is not the same, actually the installation command is missing.
```suggestion
(this is the same as `git clone https://github.com/discsim/frank.git && cd frank && pip install .`).
```

frank/geometry.py
jeffjennings: Good to note, thanks!

frank/radial_fitters.py
jeffjennings: cool, we'll do this in another PR!

PR #56, opened 2020-02-20
Fixed warning in UV binning

### LINKED ISSUE #53: RuntimeWarning in BinUVData
Traceback: 

```
/Users/morgan/miniconda3/envs/jj/lib/python3.7/site-packages/frank/useful_funcs.py:117: RuntimeWarning: invalid value encountered in sqrt
  bin_vis_err.real = np.sqrt(bin_vis_err.real - bin_vis.real**2)
```

rbooth200:
This PR fixes #53

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -124,9 +124,7 @@ class UVDataBinner(object):
         bin_vis_err[idx2].imag = \
             np.sqrt(bin_vis_err[idx2].imag - bin_vis[idx2].imag**2)
 
-
         # Use a sensible error for bins with one baseline
-        idx1 = bin_n == 1
         bin_vis_err[idx1].real = bin_vis_err[idx1].imag = \
             1 / np.sqrt(bin_wgt[idx1])
 
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -114,8 +114,16 @@ class UVDataBinner(object):
 
         bin_vis_err[idx] /= w
 
-        bin_vis_err.real = np.sqrt(bin_vis_err.real - bin_vis.real**2)
-        bin_vis_err.imag = np.sqrt(bin_vis_err.imag - bin_vis.imag**2)
+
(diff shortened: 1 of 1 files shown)


frank/utilities.py
```diff
@@ -114,11 +114,17 @@ def __init__(self, uv, V, weights, bin_width):
...
-        bin_vis_err.real = np.sqrt(bin_vis_err.real - bin_vis.real**2)
-        bin_vis_err.imag = np.sqrt(bin_vis_err.imag - bin_vis.imag**2)
 
-        # Use a sensible error for bins with one baseline
+        # Compute the variance for cells with 2 or more data points
         idx1 = bin_n == 1
+        idx2 = idx & ~idx1
+        
+        bin_vis_err[idx2].real = \
+            np.sqrt(bin_vis_err[idx2].real - bin_vis[idx2].real**2)
```
mtazzari: Hey @rbooth200 shouldn't it be:
```py
np.sqrt((bin_vis_err[idx2].real - bin_vis[idx2].real)**2)
```
Same applies to the imag part.

rbooth200: I don't think so because `bin_vis_err = sum (w * vis**2) / sum(w)`, i.e. one of these quantities has units Jy^2 and the other Jy. Does that seem correct?

I think the problem is essentially round off error.  A quick derivation:
```
var = sum (w * (vis - mean)**2) / sum(w)
var = sum( w* vis**2 - 2 * w * vis * mean + w* mean**2) / sum(w)
var = (sum (w * vis**2) -  2*mean * sum(w * vis) + mean**2 * sum(w) /) / sum(w)
```
Since `mean =  sum(w * vis) / sum(w)`
```
var = sum (w * vis**2) / sum(w) - mean**2
```

We could avoid the round-off problem by using the top expression (`var = sum (w * (vis - mean)**2) / sum(w)`, but it would mean two passes through the loop.

PR #61, opened 2020-02-21
Added variable output types

rbooth200:
This pull request implements multiple output formats for the frank fit. 

We should wait for #55 before proceeding here.

### DIFF
--- a/frank/default_parameters.json
+++ b/frank/default_parameters.json
@@ -1,7 +1,7 @@
 {
   "input_output" : {
     "uvtable_filename" : "",
-    "save_dir"	       : "",
+    "save_dir"         : "",
     "save_profile_fit" : true,
     "save_vis_fit"     : true,
     "save_uvtables"    : false,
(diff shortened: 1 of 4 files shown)


jeffjennings:
Would you mind allowing `txt` or `dat` for the ascii formats? `.dat` is common for UVTables.

rbooth200:
What format is .dat?

jeffjennings:
just an ascii format, https://help.ceda.ac.uk/article/4429-ascii-formats

rbooth200:
Ok, .dat is sometimes used for binary too. I've added it as an alias of txt

rbooth200:
You've done it now, so I wouldn't undo it - but probably we're going to get a bunch of conflicts from the changes to the formatting. It would have been best to do that as a separate PR after this and #55

jeffjennings:
my branch also has the unformatted `.json`s, so I think we might avoid conflicts

rbooth200:
Ok

jeffjennings:
should I merge?

rbooth200:
Yes

PR #64, opened 2020-02-21
Update website (not yet tutorials), update plotting as needed, other minor changes

### LINKED ISSUE #58: Update inline function calls and check notebooks in docs
Code call signatures have changed since I wrote docs. Update:

- [ ] quickstart
- [ ] tutorial: "What’s fit.py doing? The fitting procedure in detail"
- [ ] tutorial: "Running fits in a loop"
- [ ] tutorial: "Assessing a fit's uncertainty and sensitivity to hyperpriors"

jeffjennings:
Finalize the docs _except_ the tutorials. 

Some changes in the plotting scripts and calls to these, as needed to update the docs.

Only call `_ps_covariance` when it's needed.

Add `bin_edges` property to `UVDataBinner`

`"save_uvtables"    : true` by default, with the corresponding logging to tell the user these will be saved to be added in #68, and the ability to run the fit only to generate a UVTable to be added in #66 

This PR doesn't close any open issues


frank/default_parameters.json
```diff
@@ -4,7 +4,7 @@
     "save_dir"         : "",
     "save_profile_fit" : true,
     "save_vis_fit"     : true,
-    "save_uvtables"    : false,
+    "save_uvtables"    : true,
```
jeffjennings: I realize this creates two ~10 MB datafiles if they're .npz, but I expect the vast majority of users will want an output UVTable to compare with CLEAN or something else in CASA, so the default parameters should produce this.

rbooth200: These can be computed from the sol object and the uv table. Perhaps its best to add an option to fit.py that generates these files afterwards. E.g. 
Python -m frank.fit -p params --compute_vis_fit sol.obj

rbooth200: Or Python -m frank.fit -uv uvtable --compute_vis_fit sol.obj

rbooth200: This is also more flexible since you can use it to compute the vis for different UV tables if you want - might be useful for prediction / proposals

jeffjennings: That's a good idea, an additional option to just generate the uv tables without refitting (even though in most cases the fitting will be very fast). I've made an issue for this and will add the call syntax and function for it in a separate PR. I think 
`python -m frank.fit -uv uvtable --generate_uvtable sol.obj ` 
would be good. 

I do think it's still best to have `save_uvtables : true` in the default parameter file, as the majority of users, especially as they start using frank for the first time, will want the frank UVTable to compare with clean. I'll add to the logging that when this option is true, frank will be saving 2 UVTables whose size is comparable to the input UVTable. The UVTable saving is 
also the last step in the fit, so if a user killed the process at this point, everything else in the fit would already be saved.

The plotting should also have optionally produce a figure that makes: 
- the unconvolved, deprojected 2D sweep; unconvolved, reprojected 2D sweep; unconvolved, reprojected residuals using the residual UVTable (although this would require some additional functionality to make)
- the same but convolved with a beam, if the user passes in the beam
I've opened an issue for this so it can go in a separate PR.

rbooth200: OK

frank/radial_fitters.py
```diff
@@ -922,6 +923,10 @@ def MAP_spectrum(self):
     @property
     def MAP_spectrum_covariance(self):
         """Covariance matrix of the maximum a posteriori power spectrum"""
+        if self._ps_cov is None:
```
rbooth200: Did you test this? Can you confirm it works for me.

jeffjennings: I confirm it returns `MAP_spectrum_covariance` as, e.g.,

```
Message: 'FF.MAP_spectrum_covariance'
Arguments: (array([[ 1.81711780e+00,  2.11017083e-03, -1.39366320e-03, ...,
         2.91032411e-07,  3.09998364e-07,  3.31108797e-07],
       [ 2.11017083e-03,  1.81184208e+00,  1.46379719e-02, ...,
        -7.32676623e-07, -7.79869609e-07, -8.32481953e-07],
       [-1.39366320e-03,  1.46379719e-02,  1.85701394e+00, ...,
         1.05319287e-06,  1.12157499e-06,  1.19773501e-06],
       ...,
       [ 2.91032411e-07, -7.32676623e-07,  1.05319287e-06, ...,
         7.97298342e-01,  8.14643340e-01,  8.28254145e-01],
       [ 3.09998364e-07, -7.79869609e-07,  1.12157499e-06, ...,
         8.14643340e-01,  8.94548276e-01,  9.69512558e-01],
       [ 3.31108797e-07, -8.32481953e-07,  1.19773501e-06, ...,
         8.28254145e-01,  9.69512558e-01,  1.11490208e+00]]),)
```

but I haven't checked beyond that; this seems to me sufficient though.

Issue #72, opened 2020-03-04
logging in radial_fitters to give running count of iteration number


jeffjennings:

rbooth200:
I'm happy to have this in general, but I suspect that it will be difficult to provide a correct implementation. No implementation is better than a bad one.

Have you thought about how to do this in a way that will work for both the command line and the log files?

jeffjennings:
Yes the version on my branch `vis_histogram` works by printing the real-time iteration to the terminal only, and printing the final number of iterations to the single log file.

jeffjennings:
Closing: this should've been closed by PR #73, but I had a typo

Issue #70, opened 2020-03-04
add functionality for cutting data based on maximum baseline


jeffjennings:

jeffjennings:
Closing: this should've been closed by PR #73, but I had a typo

Issue #68, opened 2020-02-26
Update logging in fit.py to inform the user UVTables will be saved


jeffjennings:
And that these will be of comparable size to the input UVTable

jeffjennings:
Closing: this should've been closed by PR #73, but I had a typo

rbooth200:
Is that true? Didn't we remove the logging from that PR?

jeffjennings:
The logging is no longer in fit.py, but in the version of io.py I've opened in PR #76 . I think it easiest to leave this closed as it's deprecated and point logging things to #71.

PR #73, opened 2020-03-04
Vis histogram

### LINKED ISSUE #60: Add logging to additional scripts
To track fit iteration real-time and the slower saving operations (saving UVTables)

### LINKED ISSUE #67: Flip the import functions so standard libraries are first
Per MT's suggestion. Do this at the top of all scripts. Will take 1 minute.

jeffjennings:
Add bootstrap functions and figure; add visibility histogram plot to full figure; add functionalities to estimate and apply a correction to (mock) data weights, to normalize uv-points by an observing wavelength, to only fit data within a given baseline range; other minor additions / general cleanup to close some issues (mainly moving the logging from the highest level in fit.py to the code's internal called functions)

Fixes #67 #68 #70 #72


frank/default_parameters.json
```diff
@@ -11,7 +12,10 @@
 
   "modify_data" : {
     "cut_data"         : false,
-    "cut_baseline"     : null
+    "cut_range"        : null,
```
rbooth200: There are two parameters being used where a single one if sufficient. Also, be more explicit in the name, something like "baseline_range" would suffice

frank/default_parameters.json
```diff
@@ -11,7 +12,10 @@
...
+    "correct_weights"  : false,
+    "norm_by_wle"      : false,
+    "wle"              : null
```
rbooth200: I don't like this at all the wle should be provided in the uv table and we should detect automatically from the uv table that units to use

frank/default_parameters.json
```diff
@@ -42,5 +46,10 @@
     "force_style"      : true,
     "bin_widths"       : [1e3, 5e4],
     "iter_plot_range"  : null
+  },
+
+  "analysis" : {
+    "bootstrap"        : false,
+    "n_trials"         : null
```
rbooth200: Again, I don't like there being two not obviously linked parameters. 

```suggestion
  "bootstrap" : {
    "on"               : false,
    "n_trials"         : null
```

frank/geometry.py
```diff
@@ -22,10 +22,12 @@
    NOTE: The sign convention used here is xx.
 """
 
-from frank.constants import rad_to_arcsec, deg_to_rad
-
 import numpy as np
 from scipy.optimize import least_squares
+import logging
```
rbooth200: Remove all logging from the main library in this PR 
```suggestion
```

frank/geometry.py
```diff
@@ -216,6 +218,12 @@ def fit(self, u, v, V, weights):
         weights : array of real, size = N, unit = Jy
             Weights on the visibilities
         """
+
+        if all(x == 0 for x in (self._inc, self._PA, self._dRA, self._dDec)):
+            logging.info("      N.B.: All geometry parameters are 0 --> No geometry"
+                         " correction will be applied to the visibilities"
+                         )
+
```
rbooth200:
```suggestion
```

frank/geometry.py
```diff
@@ -304,6 +312,14 @@ def fit(self, u, v, V, weights):
...
         """
 
+        t1 = time.time()
+
+        if self._phase_centre:
+            logging.info('    Fitting Gaussian to determine geometry'
+                         ' (not fitting for phase center)')
+        else:
+            logging.info('    Fitting Gaussian to determine geometry')
+
```
rbooth200:
```suggestion
```

frank/geometry.py
```diff
@@ -312,6 +328,15 @@ def fit(self, u, v, V, weights):
...
 
+        logging.info('    Time taken for geometry %.1f sec' %
+                     (time.time() - t1))
+
+        logging.info('    Using: inc  = {:.2f} deg,\n           PA   = {:.2f} deg,\n'
+                     '           dRA  = {:.2e} mas,\n'
+                     '           dDec = {:.2e} mas'.format(self._inc, self._PA,
+                                                           self._dRA*1e3,
+                                                           self._dDec*1e3))
+
```
rbooth200:
```suggestion
```

frank/io.py
```diff
@@ -23,7 +23,7 @@
 import os
 import numpy as np
 import pickle
-
+import logging
```
rbooth200:
```suggestion
```

frank/io.py
```diff
@@ -45,7 +45,9 @@ def load_uvtable(data_file):
           Weights on the visibilities, of the form
           :math:`1 / \sigma^2`
     """
-    import os.path
+
+    logging.info('  Loading UVTable')
+
```
rbooth200:
```suggestion
```

frank/io.py
```diff
@@ -164,6 +171,8 @@ def save_fit(u, v, vis, weights, sol, prefix,
                      units={'uv': 'lambda', 'V': 'Jy'})
 
     if save_uvtables:
+        logging.info('    Saving fit and residual UVTables. N.B.: These will'
+                     ' be of comparable size to your input UVTable.')
```
rbooth200:
```suggestion
```

frank/make_figs.py
```diff
@@ -18,26 +18,30 @@
 #
 """This module generates figures for a Frankenstein fit and its diagnostics.
 """
+
+import os
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.gridspec import GridSpec
+import logging
```
rbooth200: I'm leaving the logging here because I don't see it as part of the main library, so I don't mind polluting it.

frank/parameter_descriptions.json
```diff
@@ -11,7 +12,10 @@
 
   "modify_data" : {
     "cut_data"         : "Whether to truncate the visibilities at a given maximum baseline prior to fitting",
-    "cut_baseline"     : "Maximum baseline at which visibilities are truncated"
+    "cut_range"        : "Baseline range outside of which visibilities are truncated [\\lambda]",
+    "correct_weights"  : "Whether to estimate the data's weights (as they may be unknown for mock data)",
+    "norm_by_wle"      : "Whether to normalize u and v by the observing wavelength",
+    "wle"              : "Observing wavelength [m]"
```
rbooth200: Fix these when fixing the parameters

frank/parameter_descriptions.json
```diff
@@ -42,5 +46,10 @@
     "force_style"      : "Whether to use preconfigured matplotlib rcParams in generated figures",
     "bin_widths"       : "Bin widths in which to bin the observed visibilities [\\lambda], list of float or int",
     "iter_plot_range"  : "Range of iterations to be plotted in the diagnostic figure, list of int of the form [start, stop]"
+  },
+
+  "analysis" : {
+    "bootstrap"        : "Whether to perform a bootstrap analysis on the fit",
+    "n_trials"         : "Number of trials (dataset realizations) in the bootstrap analysis"
```
rbooth200: Fix

frank/plot.py
```diff
@@ -20,15 +20,21 @@
...
 
-from frank.utilities import sweep_profile
-
 import numpy as np
 import matplotlib.pyplot as plt
 
+# Suppress some benign warnings
+import warnings
+warnings.filterwarnings('ignore', '.*compatible with tight_layout.*')
+warnings.filterwarnings('ignore', '.*handles with labels found.*')
```
rbooth200: I don't like ignoring these warnings. Can you remove the need to use tight layout by specifying the plot geometry explicitly

frank/radial_fitters.py
```diff
@@ -20,13 +21,15 @@
...
 
-from frank.hankel import DiscreteHankelTransform
-from frank.constants import rad_to_arcsec, deg_to_rad
-
 import numpy as np
 import scipy.linalg
 import scipy.sparse
 from collections import defaultdict
+import logging
+import time
```
rbooth200: Remove logging from main library in this PR

```suggestion
```

frank/radial_fitters.py
```diff
@@ -680,6 +683,9 @@ def fit(self, u, v, V, weights=1):
         # Fit geometry if needed
         self._geometry.fit(u, v, V, weights)
 
+        logging.info('  Fitting for brightness profile')
+        t1 = time.time()
+
```
rbooth200:
```suggestion
```

frank/radial_fitters.py
```diff
@@ -716,6 +722,8 @@ def fit(self, u, v, V, weights=1):
         pi_old = 0
         while (np.any(np.abs(pi - pi_old) > self._tol * pi) and
                count <= self._max_iter):
+            print('\r    Fit iteration {}'.format(count), end='', flush=True)
+
```
rbooth200: Remove from this PR 

```suggestion
```

frank/radial_fitters.py
```diff
@@ -742,6 +750,14 @@ def fit(self, u, v, V, weights=1):
...
             count += 1
 
+        if count < self._max_iter:
+            logging.info('\n    Convergence criterion met at iteration'
+                         ' {}'.format(count-1))
+        else:
+            logging.info('\n    Convergence criterion not met; fit stopped at'
+                         ' max_iter specified in your parameter file,'
+                         ' {}'.format(self._max_iter))
+
```
rbooth200: Perhaps we should raise an exception, rather than logging a warning?
```suggestion
        if count >= self._max_iter:
            raise ValueError("Reached Maximum number of iterations before converging")
```

frank/radial_fitters.py
```diff
@@ -753,6 +769,12 @@ def fit(self, u, v, V, weights=1):
         self._ps_cov = None
         self._ps_cov_params = (fit, Tij, rho)
 
+        logging.info('    Time taken to fit profile (with {:.0e} visibilities and'
+                     ' {:d} collocation points) {:.1f} sec'.format(len(u),
+                                                                   self.size,
+                                                                   time.time() - t1)
+                                                                   )
+
```
rbooth200:
```suggestion
```

frank/utilities.py
```diff
@@ -22,12 +22,12 @@
 
 import numpy as np
 from scipy.interpolate import interp1d
-
+import logging
```
rbooth200:
```suggestion
```

frank/utilities.py
```diff
@@ -168,6 +168,159 @@ def bin_edges(self):
...
+    vis_cut : array, unit = Jy
+        Visibilities in the chosen baseline range
+    weights_cut : array, unit = Jy^-2
+        Weights in the chosen baseline range
+    """
+
+    logging.info('  Cutting data outside of the minmum and maximum baselines'
+                 ' of {} and {}'
+                 ' klambda'.format(cut_range[0] / 1e3, cut_range[1] / 1e3))
+
```
rbooth200:
```suggestion
```

frank/utilities.py
```diff
@@ -168,6 +168,159 @@ def bin_edges(self):
...
+    -------
+    wcorr_estimate : float
+        Correction factor by which to adjust the weights
+
+    weights_corrected : array, unit = Jy^-2
+        Corrected weights assigned to observed visibilities, of the form
+        :math:`1 / \sigma^2`
+    """
+
+    logging.info('  Estimating, applying correction factor to the weights') # TODO: should go in utilities.py
```
rbooth200:
```suggestion
```

frank/fit.py
```diff
@@ -132,19 +132,29 @@ def parse_parameters(*args):
...
-        if plotting['iter_plot_range'] is not None:
-            err = ValueError("iter_plot_range should be 'null' (None) "
-                             "or a list specifying the start and end "
-                             "points of the range to be plotted")
+    if model['modify_data']['cut_data']:
+        if model['modify_data']['cut_range'] is not None:
+            err = ValueError("cut_range should be 'null' (None)"
+                             " or a list specifying the low and high"
+                             " baselines [unit: \\lambda] outside of which the"
+                             " data will be truncated before fitting")
```
rbooth200: Change to use a single parameter

frank/fit.py
```diff
@@ -364,145 +359,164 @@ def output_results(u, v, vis, weights, sol, iteration_diag, iter_plot_range,
...
+    Returns
+    -------
+    boot_fig : Matplotlib `.Figure` instance
+        The produced figure, including the GridSpec
+    boot_axes : Matplotlib `~.axes.Axes` class
+        The axes of the produced figure
+    """
+    profiles_bootstrap = []
+
+    for ii in range(model['analysis']['n_trials']):
```
rbooth200: Rename ii to something more informative, e.g. trial

```suggestion
    for trial in range(model['analysis']['n_trials']):
```

frank/fit.py
```diff
@@ -364,145 +359,164 @@ def output_results(u, v, vis, weights, sol, iteration_diag, iter_plot_range,
...
+        logging.info(' Bootstrap trial {} of {}'.format(ii + 1,
```
rbooth200:
```suggestion
        logging.info(' Bootstrap trial {} of {}'.format(trial + 1,
```

frank/fit.py
```diff
@@ -364,145 +359,164 @@ def output_results(u, v, vis, weights, sol, iteration_diag, iter_plot_range,
...
+    profiles_path = model['input_output']['save_prefix'] + \
+                        '_bootstrap_profiles.txt'
+    collocation_points_path = model['input_output']['save_prefix'] + \
+                                  '_bootstrap_collocation_pts.txt'
+
+    logging.info(' Bootstrap complete. Saving fitted brightness profiles and'
+                 ' the common set of collocation points')
+
+    np.savetxt(profiles_path, profiles_bootstrap)
+    np.savetxt(collocation_points_path, sol.r)
```
rbooth200: I would like these both stored in the same file. It might be worth using a .npz too, since there could be a lot of trials.

```suggestion
    np.savez(bootstrap_file, r=so.r, profiles=np.array(profiles_bootstrap))
```

frank/fit.py
```diff
@@ -364,145 +359,164 @@ def output_results(u, v, vis, weights, sol, iteration_diag, iter_plot_range,
...
+
+        utilities.draw_bootstrap_sample(u, v, vis, weights)
+
+        sol, iteration_diagnostics = perform_fit(u, v, vis, weights, geom, model)
+        profiles_bootstrap.append(sol.mean)
+
+    profiles_path = model['input_output']['save_prefix'] + \
+                        '_bootstrap_profiles.txt'
+    collocation_points_path = model['input_output']['save_prefix'] + \
+                                  '_bootstrap_collocation_pts.txt'
```
rbooth200:
```suggestion
    bootstrap_file = model['input_output']['save_prefix'] + \
                        '_bootstrap.npz'
```

frank/fit.py
```diff
@@ -364,145 +359,164 @@ def output_results(u, v, vis, weights, sol, iteration_diag, iter_plot_range,
...
+        sol, iteration_diagnostics = perform_fit(u, v, vis, weights, geom, model)
+
+        figs, axes, model = output_results(u, v, vis, weights, sol,
+                                    iteration_diagnostics, model
+                                    )
+
+        logging.info('  Updating {} with final parameters used in'
+                     ' fit'.format(param_path))
+        with open(param_path, 'w') as f:
+            json.dump(model, f, indent=4)
```
rbooth200: Why is this at the end?

frank/default_parameters.json
jeffjennings: The wle is not always printed in the table header when extracting a table in CASA (e.g., if something other than uvplot is used to extract the table).

rbooth200: I think we should just not handle the case where baseline is in meters

rbooth200: Either that, or provide a specification for a file format that frankenstein understands and that users must use.

frank/geometry.py
jeffjennings: I'm just commenting it out for now, not reverting all the changes I've made that I think should be implemented eventually. I have thought about how to do the logging and think it important that some status reports are printed even when the code is run as a module, otherwise it will be less intuitive to use and harder to debug. I'll also comment the logging calls in the main library for now.

rbooth200: I want them removed.

frank/make_figs.py
jeffjennings: In general it's not polluting, it's placing a useful functionality in the appropriate sections of the code.

frank/plot.py
jeffjennings: It's the most useful way to quickly format the figures across a range of backends used. It works well for the figures I produce, and I've made the warnings ignored strictly only these 2. Even if they were printed, they would not affect the code beyond the figure aesthetics. There won't be any missed important error messages, because only these 2 benign ones are ignored.

frank/radial_fitters.py
jeffjennings: I've already thought about and tested this, so there's no need to just remove it; I'm commenting for now.

frank/plot.py
rbooth200: Is it possible to place this in a with statement exactly where the calls are? Its best not to silence warnings globally

frank/radial_fitters.py
jeffjennings: I see your point, though in general I except fits may often reach the maximum number of iterations for datasets with properties we haven't seen yet.  I think if we want to raise an error because the fit hasn't converged:  
- we should be confident that the convergence criterion is robust 
- if we're not confident in it, we should increase the default number of iterations to some very high number. But I could see this potentially resulting in fits running much longer than needed (the convergence criterion has only been tested across a few datasets, not a wide diversity of them).

rbooth200: Raising an error is the correct default behaviour. If you want to add an option to suppress it, feel free.

frank/fit.py
jeffjennings: In `fit.py` --> `output_results` --> `if model['plotting']['diag_plot']:` --> `model['plotting']['iter_plot_range'] = iter_plot_range` gets updated based on the final actual iteration of the fit.

This is also a future-proof catchall for any other model parameters that may be updated during a fit.

rbooth200: Ok, is it worth writing a temporary one at the start, and overwriting it at the end? So the users can inspect it if the code crashes?

jeffjennings:
All previous review comments without a reply from me have been incorporated in the last 2 commits.

frank/fit.py
rbooth200: I'm not sure that, for this particular case, we want the changes written to file.

rbooth200:
Can you go over my comments?

frank/fit.py
jeffjennings: Good point, it's better not to update the parameter file in this case as the log will say how many iterations were performed anyway. 

I think it's a bit cleaner just to overwrite the same used parameter file instead of creating a temporary one used during the fit, deleting it and dumping to a final one.

rbooth200: I'm Ok with that

jeffjennings:
> Can you go over my comments?

I have just replied to your most recent one. I responded to the ones I think should be discussed further and integrated the rest in the most recent commits.

jeffjennings:
Could you now re-review?

frank/default_parameters.json
rbooth200: I still think this needs addressing.

frank/fit.py
jeffjennings: ok cool, the current version does this

frank/fit.py
```diff
@@ -203,71 +252,46 @@ def determine_geometry(u, v, vis, weights, inc, pa, dra, ddec, geometry_type,
...
-        raise ValueError("geometry_type must be one of 'known' or 'gaussian'")
 
-    logging.info('    Using: inc  = {:.2f} deg,\n           PA   = {:.2f} deg,\n'
-                 '           dRA  = {:.2e} mas,\n'
-                 '           dDec = {:.2e} mas'.format(geom.inc, geom.PA,
-                                                       geom.dRA*1e3,
-                                                       geom.dDec*1e3))
+    else:
+        raise ValueError("geometry_type in your parameter file must be one of"
+                         " 'known' or 'gaussian'")
```
rbooth200: You may want to add this logging back in here

frank/fit.py
rbooth200: Is there a temporary json file written and then overwritten?

frank/fit.py
jeffjennings: I don't follow, I already have that `raise ValueError` in the current commit.

rbooth200: You moved the logging printing the geometry to geometry.py, I've asked for it to be removed from there, so you might want to print the geometry in fit.py

jeffjennings:
> The problem with wle still needs fixing, and I want to logging comments removed.

See my comment on your initial review re: wle. It's not easily generalizable to read it in from a UVTable, as the UVTable header may not always contain it.

I'll leave the logging commented, it doesn't affect anything (there are other commented blocks in radial_fitters), and it's the solution I think should be implemented eventually, which we can discuss further.

rbooth200:
> See my comment on your initial review re: wle. It's not easily generalizable to read it in from a UVTable, as the UVTable header may not always contain it.

In which case, we should only support input in units of lambda

jeffjennings:
> > See my comment on your initial review re: wle. It's not easily generalizable to read it in from a UVTable, as the UVTable header may not always contain it.
> 
> In which case, we should only support input in units of lambda

But it's only used to normalize u and v if they're in [m], or are you thinking of a different use for wle?

rbooth200:
>But it's only used to normalize u and v if they're in [m], or are you thinking of a different use for wle?

No, I'm saying remove it. How do you multi-frequency data? Its almost certainly the case that the raw data is multi-frequency and we don't support it. I don't want frank to introduce bandwidth smearing by doing something silly.

frank/geometry.py
```diff
@@ -24,8 +24,6 @@
 
 import numpy as np
 from scipy.optimize import least_squares
-#import logging
```
jeffjennings: This isn't the way to handle a conversation. I'll now need to create a new branch with the logging changes. It would be better to just not merge this branch until we discuss further. If you run the code with the logging inline, you'll see it works as expected. So I'll make a new branch, but in the future please don't do this.

rbooth200: In future, one thing per PR

jeffjennings:
> > But it's only used to normalize u and v if they're in [m], or are you thinking of a different use for wle?
> 
> No, I'm saying remove it. How do you multi-frequency data? Its almost certainly the case that the raw data is multi-frequency and we don't support it. I don't want frank to introduce bandwidth smearing by doing something silly.
> 
> It's important that what we provide is always correct.

I agree supporting multi-frequency would be a useful extension. 'wle' was 'null' in the default parameter file, and the single wavelength case still had use. But I'll remove this now. I think it's worth keeping the function in `utilities.py` but just removing the call to it in `fit.py`.

frank/radial_fitters.py
```diff
@@ -716,6 +717,8 @@ def fit(self, u, v, V, weights=1):
         pi_old = 0
         while (np.any(np.abs(pi - pi_old) > self._tol * pi) and
                count <= self._max_iter):
+            #print('\r    Fit iteration {}'.format(count), end='', flush=True)
```
rbooth200:
```suggestion
            if verbose:
                print('\r    FrankFitter iteration {}'.format(count), end='', flush=True)
```

If you want to add it in, I think this is fine. You'd need to add verbose as a kwarg to fit

rbooth200:
I'm OK for this to go, but I've added a suggestion that you could include, if you wish

frank/radial_fitters.py
jeffjennings: cool, good point, I've updated in `radial_fitters.py` and the call in `fit.py`

frank/radial_fitters.py
```diff
@@ -717,7 +717,8 @@ def fit(self, u, v, V, weights=1):
         pi_old = 0
         while (np.any(np.abs(pi - pi_old) > self._tol * pi) and
                count <= self._max_iter):
-            #print('\r    Fit iteration {}'.format(count), end='', flush=True)
+            if verbose:
+                print('\r    Fit iteration {}'.format(count), end='', flush=True)
```
rbooth200: Best to be verbose:
```suggestion
                print('\r    FrankFitter iteration {}'.format(count), end='', flush=True)
```

jeffjennings:
Sorry for all the commits. From now on I'll do `py.test frank/tests.py` locally before pushing.

PR #75, opened 2020-03-04
Fits with non-negative brightness

rbooth200:
I've added the option to produce non-negative fits. 

The code proceeds as normal, but right at the end I am finding the best-fit non-negative brightness given the MAP power spectrum. I've added an option to the bootstrap to use this instead of the mean, if desired.

### DIFF
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -21,7 +21,6 @@
    and output results. Alternatively a custom parameter file can be provided.
 """
 
-from frank import io, geometry, make_figs, radial_fitters, utilities
 import os
 import sys
 import time
@@ -31,11 +30,8 @@ import numpy as np
 import logging
 
 import frank
-<<<<<<< HEAD
 from frank import io, geometry, make_figs, radial_fitters, utilities
 
-=======
->>>>>>> Added non-negative option
 frank_path = os.path.dirname(frank.__file__)
 
 
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -21,6 +21,7 @@
    and output results. Alternatively a custom parameter file can be provided.
 """
 
+from frank import io, geometry, make_figs, radial_fitters, utilities
 import os
 import sys
 import time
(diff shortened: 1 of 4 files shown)


frank/fit.py
```diff
@@ -21,6 +21,7 @@
    and output results. Alternatively a custom parameter file can be provided.
 """
 
+from frank import io, geometry, make_figs, radial_fitters, utilities
```
jeffjennings: Sorry can you change this back? Marco specifically asked that all frank imports go after standard python imports.

rbooth200: Sure

jeffjennings:
Sorry I see that by leaving a comment I opened a review; do you want me to review now or wait?

rbooth200:
Yes review now

frank/parameter_descriptions.json
```diff
@@ -46,6 +46,8 @@
   },
 
   "analysis" : {
-    "bootstrap_ntrials": "Number of trials (dataset realizations) to perform in a bootstrap analysis"
+    "bootstrap_ntrials": "Number of trials (dataset realizations) to perform in a bootstrap analysis",
+    "bootstrap_non_negative" : "Determines whether the mean (if false) or best-fit non-negative solution in used in the bootstrap"
```
jeffjennings: Very minor, but can you update to 

"bootstrap_non_negative" : "Whether the mean (if false) or best fit nonnegative solution in used in the bootstrap"

rbooth200: Done

frank/radial_fitters.py
```diff
@@ -745,7 +757,10 @@ def fit(self, u, v, V, weights=1, verbose=False):
                 self._iteration_diagnostics['mean'].append(fit.mean)
 
             count += 1
-
+            
+        if verbose:
+            print()
```
jeffjennings: I don't need this to print the line at 732, do you?

rbooth200: Its there so add a new line after the "    FrankFitter iteration XXX" for the last iteration

rbooth200:
See comments

frank/radial_fitters.py
jeffjennings: The print statement uses flush so that it reprints on the same line at each iteration.

rbooth200: Yes, but the line following it will begin on the same line as the iteration statement unless you add that print

jeffjennings: ok

PR #76, opened 2020-03-04
logging messages

### LINKED ISSUE #71: integrate logging across scripts
And check logging behavior if using frank as module (if you don't run fit.py, you'd have to instantiate the logfile somewhere else).

jeffjennings:
Integrate logging throughout code so that useful status reports and diagnostics are printed when running the code as a module (not just when using fit.py)

Fixes #71


rbooth200:
Before you waste time on this, I really think less is more with these kinds of prints

jeffjennings:
> Before you waste time on this, I really think less is more with these kinds of prints

I wouldn't call it a waste, it's important functionality. But I don't really see the issue; these are the same prints that were in fit.py before. I can wrap them all in an 'if verbose' if that's better.

rbooth200:
I'm ok with most of that logging, but I think most of it (80+% belongs in fit.py)

jeffjennings:
> I'm ok with most of that logging, but I think most of it (80+% belongs in fit.py)

Except that if you run the code as a module, you wno't see anything in fit.py. I think the best compromise is to put it logging statements throughout the code that are only printed if `verbose=True`. 

In general I think it makes more sense to have a statement like "Saving fit" in the function where the fit is actually saved, not in the line above that call in fit.py, in part because the order/structure of calls in fit.py could evolve, which introduces a potential for confusion.

rbooth200:
> The main thing for me is that if you run the code as a module, you won't see anything in fit.py. I think the best compromise is to put logging statements throughout the code that are only printed if verbose=True.

I don't like having verbose=True everywhere either. I acutally think its better to try to control this using the logging levels

rbooth200:
> In general I think it makes more sense to have a statement like "Saving fit" in the function where the fit is actually saved, not in the line above that call in fit.py, in part because the order/structure of calls in fit.py could evolve, which introduces a potential for confusion.

On the contrary, it doesn't make sense for a function that only saves a fit to print "saving fit". I called that function, so I know its going to save the fit. logging statements like that are only useful if you're calling code that does multiple things

jeffjennings:
It's not on the contrary; as another example, it makes sense for a function
that iterates the fit to print *Convergence criterion met at iteration {}*
instead of storing that number and passing it to a wrapper function just to
print it, just as it makes sense for *print('\r    FrankFitter iteration
{}'.format(count), end='', flush=True) *to be internal to the function
that's iterating the fit.

The main motivations for 'embedding' logging statements are:
- debugging, preventing confusion with tracebacks when print statements are
external to the functions that are actually doing what the print statement
says, and
- giving some status on the fit to users running the code as a module. This
will also make tracing errors easier for them.

I have the logging level set in *fit.py*.






*    logging.basicConfig(level=logging.INFO,
format='%(message)s',                        handlers=[
        logging.FileHandler(log_path, mode='w'),
logging.StreamHandler()]                        )*

I can set this level in the other scripts to be the same value, '*INFO',*
if verbose=True, else a higher value, so then there's only 1 verbose check
in each script, and if verbose=False, the logging statements that update
the user on the status of the fit won't be printed.

frank/geometry.py
```diff
@@ -217,6 +219,11 @@ def fit(self, u, v, V, weights):
             Weights on the visibilities
         """
 
+        if all(x == 0 for x in (self._inc, self._PA, self._dRA, self._dDec)):
+            logging.info("      N.B.: All geometry parameters are 0 --> No geometry"
+                         " correction will be applied to the visibilities"
+                         )
+
```
rbooth200: I'm against having this here - you can't get all zeros by accident, you have to have passed them in. If you want to check in fit.py, fine, but not here.

frank/geometry.py
```diff
@@ -305,6 +312,14 @@ def fit(self, u, v, V, weights):
...
         """
 
+        t1 = time.time()
+
+        if self._phase_centre:
+            logging.info('    Fitting Gaussian to determine geometry'
+                         ' (not fitting for phase center)')
+        else:
+            logging.info('    Fitting Gaussian to determine geometry')
+
```
rbooth200: I think this is fine

frank/geometry.py
```diff
@@ -313,6 +328,15 @@ def fit(self, u, v, V, weights):
...
 
+        logging.info('    Time taken for geometry %.1f sec' %
+                     (time.time() - t1))
+
+        logging.info('    Using: inc  = {:.2f} deg,\n           PA   = {:.2f} deg,\n'
+                     '           dRA  = {:.2e} mas,\n'
+                     '           dDec = {:.2e} mas'.format(self._inc, self._PA,
+                                                           self._dRA*1e3,
+                                                           self._dDec*1e3))
+
```
rbooth200: I think the timing and printing of the geometry belongs in fit.py There is no need to print the results by default since they are stored and easily accessible in future code. 

If you *really* want to make it easy for a user to see the geometry values, add a __str__ method that returns this as a string. Then if you call print(geometry) you'll get a print out of the geometry.

frank/io.py
```diff
@@ -44,6 +45,9 @@ def load_uvtable(data_file):
           Weights on the visibilities, of the form
           :math:`1 / \sigma^2`
     """
+
+    logging.info('  Loading UVTable')
+
```
rbooth200: I genuinely think this is unnecessary. If you *really* want to log this function, do it at the debug level, not info

frank/io.py
```diff
@@ -138,6 +142,7 @@ def save_fit(u, v, vis, weights, sol, prefix, save_solution=True,
         File format in which to save the fit's output UVTable(s)
     """
 
+    logging.info('  Saving results')
```
rbooth200: This is fine, but could be more explicit about what 'results' are

frank/radial_fitters.py
```diff
@@ -692,6 +694,9 @@ def fit(self, u, v, V, weights=1, verbose=False):
         # Fit geometry if needed
         self._geometry.fit(u, v, V, weights)
 
+        logging.info('  Fitting for brightness profile')
+        t1 = time.time()
```
rbooth200: Please moving timing to fit.py

frank/radial_fitters.py
```diff
@@ -757,10 +762,18 @@ def fit(self, u, v, V, weights=1, verbose=False):
...
             print()
             
+        if count < self._max_iter:
+            logging.info('    Convergence criterion met at iteration'
+                         ' {}'.format(count-1))
+        else:
+            logging.info('    Convergence criterion not met; fit stopped at'
+                         ' max_iter specified in your parameter file,'
+                         ' {}'.format(self._max_iter))
+
```
rbooth200: I'm OK with this

frank/radial_fitters.py
```diff
@@ -772,6 +785,12 @@ def fit(self, u, v, V, weights=1, verbose=False):
         self._ps_cov = None
         self._ps_cov_params = (fit, Tij, rho)
 
+        logging.info('    Time taken to fit profile (with {:.0e} visibilities and'
+                     ' {:d} collocation points) {:.1f} sec'.format(len(u),
+                                                                   self.size,
+                                                                   time.time() - t1)
+                                                                   )
+
```
rbooth200: Move to fit.py

frank/utilities.py
```diff
@@ -185,6 +186,9 @@ def normalize_uv(u, v, wle):
         u and v coordinates normalized by observing wavelength
     """
 
+    logging.info('  Normalizing u and v by observing wavelength of'
+                 ' {} m'.format(wle))
+
```
rbooth200: I genuinely think this is unnecessary. If you *really* want to log this function, do it at the debug level, not info

frank/utilities.py
```diff
@@ -218,6 +222,10 @@ def cut_data_by_baseline(u, v, vis, weights, cut_range):
         Weights in the chosen baseline range
     """
 
+    logging.info('  Cutting data outside of the minmum and maximum baselines'
+                 ' of {} and {}'
+                 ' klambda'.format(cut_range[0] / 1e3, cut_range[1] / 1e3))
+
```
rbooth200: I genuinely think this is unnecessary. If you *really* want to log this function, do it at the debug level, not info

frank/utilities.py
```diff
@@ -259,6 +267,8 @@ def apply_correction_to_weights(u, v, ReV, weights, nbins=300): # TODO: should c
         :math:`1 / \sigma^2`
     """
 
+    logging.info('  Estimating, applying correction factor to visibility weights') # TODO: should go in utilities.py
+
```
rbooth200: This should be left here, or if moved to utlilities.py, make it debug level

frank/radial_fitters.py
```diff
@@ -692,6 +694,9 @@ def fit(self, u, v, V, weights=1, verbose=False):
         # Fit geometry if needed
         self._geometry.fit(u, v, V, weights)
 
+        logging.info('  Fitting for brightness profile')
+        t1 = time.time()
+
         # Project the data to the signal space
         self._build_matrices(u, v, V, weights)
```
rbooth200: For consistency, I think build matrices should be logged, it can be an expensive function call - equiavelent to ~ 100 iterations, which are logged.

frank/geometry.py
jeffjennings: Moved to fit.py

frank/geometry.py
jeffjennings: I moved these to fit.py and am leaving printing of the used values as default there because if running from the terminal, the point is to give the user a quick sense of whether the geometry values are reasonable, not to ask them to wait for the fit to end and then open a python terminal to load in and print the geometry values.

frank/io.py
jeffjennings: I don't see this as unnecessary but as a useful status report to the user. I'd prefer to leave at the info level.

frank/geometry.py
rbooth200: I'm ok with them being default if they are in fit.py, but not here.

frank/utilities.py
jeffjennings: I don't see this as unnecessary but as a useful status report to the user. I'd prefer to leave at the info level.

frank/utilities.py
jeffjennings: I don't see this as unnecessary but as a useful status report to the user. I'd prefer to leave at the info level.

rbooth200:
If you want to keep those statements that I flagged as debug, move them to fit.py

frank/io.py
jeffjennings: Changed to 'Saving fit result datafiles' ; the exact datafiles depend on the parameter file. A separate logging statement is used when the UVTables are saved.

frank/radial_fitters.py
jeffjennings: Moved to fit.py

frank/radial_fitters.py
jeffjennings: Moved

frank/radial_fitters.py
jeffjennings: If not using `time` in radial_fitters.py, this would require modifying the syntax to call `FrankFitter` so that `_build_matrices` could be called separately. I'd prefer to avoid this.

rbooth200: I don't quite understand? I'm just suggesting add an extra logging message inside _build_matrices

rbooth200: I've added a commit showing you want I mean

frank/io.py
rbooth200: The reason that I don't think this is useful is that it is a simple function call. It only does one thing - load a UV table - and you are expecting that when you call the function. So it doesn't need reporting every time you call it. 

As an optional thing for debugging I can see that it would be useful. I can also see it as a useful progress report in fit.py - but if that's what its being used for then it should be there not here. I just don't see it as a useful status report otherwise.

frank/utilities.py
rbooth200: I feel the same about this one as load_uvtable

frank/utilities.py
rbooth200: I feel the same about this one as load_uvtable

jeffjennings:
I removed the 'unconstrained' and 'constrained' print statements as they'd need more detail to be clear and will probably serve to confuse people who are just running a standard frank fit from the terminal. 

I removed the timing statements from radial_fitters; they're back in fit.py as you requested. To be consistent, If they aren't in geometry.py, they shouldn't be in radial_fitters.py.

rbooth200:
> I removed the timing statements from radial_fitters; they're back in fit.py as you requested. To be consistent, If they aren't in geometry.py, they shouldn't be in radial_fitters.py.

Great, thats what I was asking for there

jeffjennings:
To me this is ready for final review / to be merged.

frank/utilities.py
jeffjennings: Moved to fit.py

frank/utilities.py
jeffjennings: Moved to fit.py

frank/utilities.py
jeffjennings: Moved to fit.py

frank/radial_fitters.py
jeffjennings: I kept the print statement that the matrices are being built, removed `time` calls

frank/io.py
jeffjennings: Moved to fit.py

frank/geometry.py
jeffjennings: Moved to fit.py

frank/radial_fitters.py
rbooth200: I thought it made sense to leave line 697 in the fit method. Do you agree?

jeffjennings: Now that almost all logging statements are back in fit.py, I think it's consistent to have it in fit.py, not radial_fitters.py.

rbooth200: The reason I disagree is that its currently possible/common to get internal logging messages from inside the .fit method without reporting that your actually doing the fit. Surely that doesn't make sense?

jeffjennings: As you said, if you're using frank as a module, you'll have to call FrankFitter directly, so you'll know you're running a fit.

jeffjennings: It's the same logging behavior as in geometry.py

rbooth200: It's not though is it? FitGeometryGaussian always gives a message that it's doing a Gaussian fit to the gemetry

jeffjennings: It does, but FixedGeometry doesn't print anything, so if you passed that when using the code as a module, you wouldn't get a notification the visibilities are being deprojected.

Similarly in io.py you get a notification that the UVTables are being saved, but the notification that the fit results are being saved is in fit.py as you requested.

In all 3 cases the behavior should be the same, either all 'parent' notifications in fit.py, or in the respective scripts.

rbooth200:
Have you tried this using frank as a library? None of the logging statements get printed

jeffjennings:
That's #71; you'd have to instantiate the logging config somewhere outside fit.py for anything to be printed. I detached this PR from #71, which I can address in a future PR (not quite sure where the best place to instantiate would be).

jeffjennings:
> Have you tried this using frank as a library? None of the logging statements get printed

That's #71; you'd have to instantiate the logging config somewhere outside fit.py for anything to be printed. I detached this PR from #71, which I can address in a future PR (not quite sure where the best place to instantiate would be).

rbooth200:
I think #71 needs to be fixed before we can merge this

jeffjennings:
> I think #71 needs to be fixed before we can merge this

I'm not sure where the instantiation (below) should go for that case; even in io.load_uvtable isn't ideal because maybe someone just loads a uvtable with genfromtxt without calling load_uvtable before performing a fit.

Do you have any ideas?

    log_path = save_prefix + '_frank_fit.log'
    logging.basicConfig(level=logging.INFO,
                        format='%(message)s',
                        handlers=[
                            logging.FileHandler(log_path, mode='w'),
                            logging.StreamHandler()]
                        )

rbooth200:
To be honest, I'd be content if we didn't emit any logs when using it as a library - that just means finding a better solution for printing the iteration count

rbooth200:
We could provide an interface for "turning on logging" so that users can have it if they want.

jeffjennings:
> To be honest, I'd be content if we didn't emit any logs when using it as a library - that just means finding a better solution for printing the iteration count

That call should still work regardless of logging because it uses `print` explicitly...not a terrible thing to show how quickly the fit is progressing for larger datasets. That's the 1 part of the code that could just seem like it's hanging otherwise.

jeffjennings:
> We could provide an interface for "turning on logging" so that users can have it if they want.

I'd just leave everything as it is for now. I see that it makes sense to provide the most 'support' for fits run from the terminal, while if someone's running the code as a module, they have a lot more control, and it's easier to debug. I wanted to include print statements even when using the code as a module, but I don't want to spend any more time on this right now.

jeffjennings:
I'd prefer to just merge this and leave #71 open; this PR works in spite of #71 and most importantly has the logging for calls to the code from the terminal.

rbooth200:
I've added a way that users can turn on logging as desired.

jeffjennings:
> I've added a way that users can turn on logging as desired.

ok cool thank you, I've relinked this PR to #71 to close it when merged

rbooth200:
I'm just looking into how to handle the fit iterations properly

rbooth200:
Ok, so I'm now automatically detecting whether to print iterations from the logging module

rbooth200:
Now that we have logging off by default when using frank as a library, but an option to turn it back on I propose to move some of the logging calls back into the main library. I'll do this now...

jeffjennings:
give me a minute before merging...

jeffjennings:
All good from my end

Issue #78, opened 2020-03-04
Frank is slower with multithreading enabled


rbooth200:
By default, modern implementations of numpy/scipy use OPENBLAS or MKL, which use multiple threads for evaluating the matrix operations. In frank, this actually makes the code slower - at least with OPENBLAS

There is however an opportunity to exploit parallelism when doing bootstrapping, which is currently missed (Its necessary to take care with numpy's random numbers though, which are not threadsafe')

rbooth200:
Fixed in #79

PR #79, opened 2020-03-04
disabled threading in numpy/scipy

rbooth200:
I've added a simple hack to take care of #78. It only works when using frank as a library

Is this too bad? There aren't many other options, apart from https://github.com/joblib/threadpoolctl

### DIFF
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -33,7 +33,9 @@ def _check_and_warn_if_parallel():
     if num_threads > 1:
         logging.warning("WARNING: You are running frank with "
                         "OMP_NUM_THREADS={}.".format(num_threads) +
-                        "The code will likely run faster on a single thread.")
+                        "The code will likely run faster on a single thread.\n"
+                        "Use unset OMP_NUM_THREADS or export OMP_NUM_THREADS=1"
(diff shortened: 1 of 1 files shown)


frank/fit.py
```diff
@@ -25,6 +25,12 @@
 import sys
 import time
 import json
+
+# Force frank to run on a single thread if we are using it as a library
+if __name__ == "__main__":
+    if 'OMP_NUM_THREADS' not in os.environ:
```
mtazzari:
```suggestion

```

frank/fit.py
```diff
@@ -25,6 +25,12 @@
...
+        os.environ['OMP_NUM_THREADS'] = '1'
```
mtazzari:
```suggestion
    os.environ.get('OMP_NUM_THREADS', '1')
```

rbooth200:
Can you explain these changes? I'm wondering what the default behaviour should be. If not specified,I think its reasonable to set it to 1. If it is specified, maybe we should just print a warning?

rbooth200:
I've added a warning message

rbooth200:
Better?

PR #87, opened 2020-03-20
URGENT: pre-release README warning

jeffjennings:
This PR quickly updates the README to be more direct that the code is still under construction (pre-v1.0). I ask that this be merged into master now so that those who find the repo understand the code isn't fully public yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 </p>
 -->
 
-You stumble through the vacant laboratory, candelabra in hand, and suddenly _**gasp!**_ your goblet echoes against the aged stone floor. You've discovered Frankenstein before his release! Imagine for a moment the _terror_, the _horror_, of a monster not yet complete! Please check back soon (hopefully in mid/late March), or if you haven't the time to wait, [email Jeff](mailto:jmj51@ast.cam.ac.uk) to begin using the code sooner. Thanks!
+**WARNING** Though you've found this repo, the code isn't quite released yet! v1.0 will be concurrent with the arXiv posting of the code's methods paper (we aim for April). Please [email Jeff](mailto:jmj51@ast.cam.ac.uk) if you'd like to use the code before then. Thanks!
 
 <p align="center">
   <img width = "800" src="https://4.bp.blogspot.com/-pd5FC25Qv1E/TtkaQcZ2wgI/AAAAAAAAB4k/5AMux-W14T4/s1600/34-crop-horz.jpg"/>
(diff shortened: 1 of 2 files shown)


mtazzari:
I'm fine with the README changes but I'd caution about changing the default Rout as well. 
3" seems more conservative than 2" and I am in favor of it, unless the change causes side effects at runtime (e.g. fit, plots change behaviour..).
Just make sure with @rbooth200 that it is fine to change the default.

rbooth200:
I'd like more justification for 3" instead of 2". One of the manuscript or the default needs to be changed to be consistent, but I'm not sure why the default is being changed. How many discs are bigger than 2"?

jeffjennings:
Thanks Marco. Yes Richard and I have both been using 3.0" as default, and
it's stable. He's seen this and is good with it.

rbooth200:
I haven't been using anything as a default, but it is definitely better to err large rather than small and I think 100 points per arcsec is a nice default resolution (although maybe slightly low).

mtazzari:
Exactly, that is my point. 0.01" can be only a factor 2.5x smaller than the beam for DSHARP-like obs.
In CLEAN, you would choose 1/3-1/5.
Is it feasible 200 points per arcsec, i.e. 0.005"? This should cover pretty much all observations.

rbooth200:
Yeah 200 points per arsec is better for AS 209. It's take 8x longer if we push it to ~600 bins though. I think 300 + 2" is maybe a better choice

jeffjennings:
300 + 2" is fine. Happy to give a sense of the minimal 'resolution' that
the fit should apply when the user chooses Rout and nbins.

rbooth200:
I think we should just add a comment to the docs that people should check this.

jeffjennings:
> I think we should just add a comment to the docs that people should check this.

yes that's what I mean, I'll add to thee quickstart.

jeffjennings:
I've opened #88 to address adding an explanation of suggested ranges for hyperprior values to the docs. That can go in a later PR; I've changed the default `Rout` back to 2". This PR for me is ready to merge.

mtazzari:
Great. I suggest you do the change to the quickstart within this PR so it is self-contained.

mtazzari:
Oh ok, forget my last comment. I didn't see you pushed the new commit. Fine to merge then

jeffjennings:
I suggest keeping it separate because it won't be a single sentence; it will need to be a dedicated page that succinctly discusses the suggested choices/ranges for all hyperpriors and demonstrates how to overplot fits to the same dataset under multiple hyperprior choices.

jeffjennings:
Great, sounds good!

PR #94, opened 2020-04-02
Fix reprojection bug

### LINKED ISSUE #93: Reprojection bug
In `geometry.deproject`, when reprojecting frank visibilities (to match the u,v coordinates in an input UVTable), the following change is needed to reproject back to the correct coordinates. Note the syntax here is based on that in the `reprojection` branch.

```
    if inverse:
        up = u * cos_t + v * sin_t
        vp = -u * sin_t + v * cos_t
        up /= np.cos(inc)
        return up, vp
```

should be reordered as

```
    if inverse:
        u /= np.cos(inc)
        up = u * cos_t + v * sin_t
        vp = -u * sin_t + v * cos_t
        return up, vp
```

Additionally `geometery.SourceGeometry.undo_correction` is currently returning the input visibilities `V` rather than the phase-unshifted visibilities `vis`:
https://github.com/discsim/frank/blob/1f17592ebcda215f5ee46901cd176efd287bef95/frank/geometry.py#L192-L195.

jeffjennings:
In `geometry.deproject`, the ordering of the code needs to be amended as detailed in #93, and the return in `geometery.SourceGeometry.undo_correction` needs to be amended as detailed in #93, to correctly reproject the u,v coordinates and phase-unshift the visibilities. This PR makes these changes (verified as correct with multiple datasets). 

This PR also:
(1) updates the syntax in a few places in `geometry.py` so that it's consistent between functions, and so that function docstrings match the input and output variables.
(2) updates `radial_fitters._HankelRegressor.predict` to add the option to return the reprojected u and v in addition to the corresponding V. The default (and existing) behavior is just to return V, consistent with the `return` in `predict_deprojected`.
(3) following from (2), the PR updates `io.save_fit` to save the frank fit and residual UVTables as reprojected and phase-unshifted (this motivates, by convenience, adding the optional `return` in (2)). This prevents frank from saving fit and residual UVTables that always yield an image produced in CASA with the source centered at the origin, even if the source in the input UVTable isn't centered (which would notably make  the resulting residual image incorrect). 

Fixes #93

### DIFF
--- a/frank/geometry.py
+++ b/frank/geometry.py
@@ -104,15 +104,15 @@ def deproject(u, v, inc, PA, inverse=False):
     sin_t = np.sin(PA)
 
     if inverse:
+        sin_t *= -1
         u /= np.cos(inc)
-        up = u * cos_t + v * sin_t
-        vp = -u * sin_t + v * cos_t
-        return up, vp
 
     up = u * cos_t - v * sin_t
     vp = u * sin_t + v * cos_t
+
     #   Deproject
-    up *= np.cos(inc)
+    if not inverse:
+        up *= np.cos(inc)
 
     return up, vp
 
--- a/frank/geometry.py
+++ b/frank/geometry.py
@@ -49,7 +49,9 @@ def apply_phase_shift(u, v, V, dRA, dDec, inverse=False):
     dDec : float, unit = arcsec
         Phase shift in declination.
         NOTE: The sign convention is xx
-
+    inverse : bool, default=False
+        If True, the visibilities are uncentered (the phase shift undone)
+        rather than centered
     Returns
     -------
     shifted_vis : array of real, size = N, unit = Jy
@@ -61,7 +63,12 @@ def apply_phase_shift(u, v, V, dRA, dDec, inverse=False):
 
     phi = u * dRA + v * dDec
 
-    return V * (np.cos(phi) + 1j * np.sin(phi))
+    if inverse:
+        shifted_vis = V / (np.cos(phi) + 1j * np.sin(phi))
+    else:
+        shifted_vis = V * (np.cos(phi) + 1j * np.sin(phi))
+
+    return shifted_vis
 
 
 def deproject(u, v, inc, PA, inverse=False):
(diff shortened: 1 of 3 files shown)


frank/geometry.py
```diff
@@ -97,16 +104,15 @@ def deproject(u, v, inc, PA, inverse=False):
     sin_t = np.sin(PA)
 
     if inverse:
-        sin_t *= -1
+        u /= np.cos(inc)
+        up = u * cos_t + v * sin_t
+        vp = -u * sin_t + v * cos_t
+        return up, vp
```
rbooth200, line 110:
```suggestion
        sin_t *= -1
        u /= np.cos(inc)
```
The minimum changes needed (re-uses as much code as possible)

frank/geometry.py
```diff
@@ -97,16 +104,15 @@ def deproject(u, v, inc, PA, inverse=False):
...
     up = u * cos_t - v * sin_t
     vp = u * sin_t + v * cos_t
-
-    # Deproject
-    if inverse:
-        up /= np.cos(inc)
-    else:
-        up *= np.cos(inc)
+    #   Deproject
+    up *= np.cos(inc)
```
rbooth200, line 115:
```suggestion
    #   Deproject
    if not inverse:
        up *= np.cos(inc)
```

jeffjennings: Done in 85bb627

frank/geometry.py
jeffjennings: Done in 85bb627

> Hi Jeff, well spotted.
> 
> I've requested a few small changes to geometry.py to keep the old code where possible because fewer changes are generally better. I'm fine with the rest of the changes in geometry.py. I'm not sure why the private function _fit_gemeotry_gaussian needed changing, but the code is no worse so I'm happy to leave it.
> 
> However, the changes to radial_fitters.py are wrong and need to be rolled back. The code originally computed the deprojected and phase corrected visibilities, so that if the source was off centre in the original data it will also be in the predicted data, as desired. I also don't think this code alters the u and v points passed in. It creates _new_ variables, u and v, that are the deprojected variables. This essentially "hides" the original u and v variables in the function, not overwriting them. The original u v and points (consistent with the CASA vis) should be unchanged.
> 
> Similarly, this means that the changes in io.py should be rolled back too.

Great, thanks for reviewing so quickly. I made all your changes in `geometry.py`. I also reverted my changes in `radial_fitters.py` and `io.py` (I don't think they weren't wrong, in that they gave the same result as the version in master, but they were thus redundant).

rbooth200: Ok, I agree just redundant

frank/radial_fitters.py
```diff
@@ -319,9 +319,7 @@ def predict(self, u, v, I=None, geometry=None, block_size=10**5):
         if geometry is not None:
             V *= np.cos(geometry.inc * deg_to_rad)
 
-        # Undo phase centering
         _, _, V = geometry.undo_correction(u, v, V)
```
rbooth200, line 322:
```suggestion
        # Undo phase centering
        _, _, V = geometry.undo_correction(u, v, V)
        
```

rbooth200:
One more small change - I added a comment back in

jeffjennings:
> One more small change - I added a comment back in

Done in 77a6572

PR #99, opened 2020-04-07
Sanity check on q

### LINKED ISSUE #85: bad parameter choices warning
Add a warning if, e.g., `hyperpriors : n` is too small for `hyperpriors : rout` (because in this case the fit can be bad without it being immediately clear why). This should be based on whether the last collocation point is at shorter baseline than the largest baseline in the data.

jeffjennings:
This PR raises a ValueError in `radial_fitters.FrankFitter` if `q[-1]` is shorter than the longest deprojected baseline in the dataset. This should help to prevent a user choosing too small `N` for a given Rmax, as in many such instances the fit can still run 'successfully,' but the result is clearly wrong, yet it's not immediately obvious why.

I also updated the `tests.test_frank_fitter` test to use large enough `N` to not raise this ValueError.

The PR also updates the default `FrankFitter.max_iter` from 1500 --> 2000, and the default `store_iteration_diagnostics` from True --> False.

Fixes #85


jeffjennings:
> This looks good, but I have a couple of comments:
> 
> * I think it would be good to add an option (maybe to the  __init__ method) to turn off these checks

Cool, I've added in e9824ae.

> * Should we also check that q[0] is not too small either? If q[0] << min(uv) then you can get a situation where the prior pulls the brightness at q[0] to zero and the frank fit will have low total flux. For this case should we consider q[0] < min(uv) or 2*q[0] < min(uv) to give some tolerance/freedom?

I think this can only happen in practice if `Rmax` is _way_ too big; e.g. for DR Tau, it took Rmax = 6.5" for q[0] to be less than min(uv).  I'm also not sure this is the same issue as  q[-1] being less than max(uv); a user might want to extrapolate the fit to somewhat lower baseline than that in the data, e.g., for extended sources (galaxies) whose size is larger than the shortest baseline. But I appreciate we don't want the total flux to be biased low. I've added the case 1*q[0] < min(uv) as a warning and kept the q[-1] < max(uv) as an error (see below) in e9824ae. I've kept 1*q[0] < min(uv)  as always checked because it just raises a warning, not an error.

> * Should these be warning messages or raising errors? Is the fit _always_ bad if they are violated?

If q[-1] is almost as large as the longest baseline, the fit can still be reasonable. Though from what I've seen, the fit can be unpredictable when q[-1] < max(uv); it's easy to get somewhat stochastic behavior in how the fit walks off the data when q[-1] < max(uv). Since we have an option to cut the uv distribution before fitting, I think that's the more robust approach if a user wants to fit to some shorter baseline. Otherwise for the sake of more stable results and more easily diagnosable fit failures, I think it's best to enforce this as an error. Plus since this can now be disabled by passing in `check_maxq=False`, if someone really wants to ignore it, they can. I've updated the error message to additionally point to `modify_data: baseline_range` in e9824ae.

rbooth200:
OK, I'm fine with using an error.

Regarding q[0], I have seen a case in bjnorfolk's data where this is an issue at 1-2"

jeffjennings:
> OK, I'm fine with using an error.
> 
> Regarding q[0], I have seen a case in bjnorfolk's data where this is an issue at 1-2"

hmm, ok I haven't seen that in any ALMA datasets at least; I didn't expect it. If you're ok to leave the q[0] case as a warning, I think this is ready to merge.

rbooth200:
Can you put both the min/max under the if check_maxq (and maybe rename the variable now)?

jeffjennings:
Now done in 9fceb93

rbooth200:
Sorry, one more thing that I should have spotted earlier. I think this should happen after the data are de-projected because the effect of de-projection is to increase the effective uv distance. 

I'm going to push a suggested implementation

rbooth200:
Ok one more question. I was surprised that this change hasn't caused the tests to fail. Do you understand why? Is 60 bins enough to fit the AS 209 dataset?

rbooth200:
Ok, got it. I'm fine with this now if you are.

jeffjennings:
cool! (yes I updated the expected array for AS 209 in the tests to that obtained with 60 bins (it was previously 20))

Issue #98, opened 2020-04-07
Estimation of weights


rbooth200:
The functions for estimating the weights are wrong unless the weights passed in are 1 to start with. Also, we should probably include the imaginary part in estimating the weights

rbooth200:
A function like this one:
```
def estimate_baseline_dependent_weight(q, V, bin_width):
  
    uvBin = UVDataBinner(q, V, np.ones_like(q), bin_width)
    var = 0.5*(uvBin.error.real**2 + uvBin.error.imag**2) * uvBin.bin_counts

    weights = np.full_like(q, np.nan)
    left, right = uvBin.bin_edges
    for i, [l,r] in enumerate(zip(left, right)):
        idx = (q >= l) & (q < r)
        weights[idx] = 1/var[i]

    assert np.all(~np.isnan(weights)), "Weights needed for all data points"
    return weights
```
Could be used to estimate the weights.

jeffjennings:
Do you want to just replace `io.apply_correction_to_weights` with this?

rbooth200:
Maybe, but I'm not totally convinced yet. I have a few questions:

- Should we add an option for computing a single estimate (e.g. the median) as well as a baseline dependent estimate?
- Should include the option of binning in log-space?
- Should we (optionally) use the information from existing weights in the estimation?

jeffjennings:
> Maybe, but I'm not totally convinced yet. I have a few questions:
> 
> * Should we add an option for computing a single estimate (e.g. the median) as well as a baseline dependent estimate?

That's a good idea, could be useful for mock data, 
```
def estimate_baseline_dependent_weight(q, V, bin_width, point_estimate=False):
  
    uvBin = UVDataBinner(q, V, np.ones_like(q), bin_width)
    var = 0.5*(uvBin.error.real**2 + uvBin.error.imag**2) * uvBin.bin_counts

    weights = np.full_like(q, np.nan)
    left, right = uvBin.bin_edges
    for i, [l,r] in enumerate(zip(left, right)):
        idx = (q >= l) & (q < r)
        weights[idx] = 1/var[i]

    assert np.all(~np.isnan(weights)), "Weights needed for all data points"

    if point_estimate:
        return np.median(weights)

    return weights
```

> * Should include the option of binning in log-space?
I think this is also a good idea, we could just add the option in `UVDataBinner` and then the above code for `estimate_baseline_dependent_weight` need not changee.

> * Should we (optionally) use the information from existing weights in the estimation?
[...]

PR #103, opened 2020-04-08
Added new weight estimation facility

rbooth200:
This updates the way the weights are determined as discussed in #98. 

I had to make some cosmetic changes to UVDataBinner to get this to work, plus a small bug-fix.

I also decided against doing something too sophisticated, since its probably difficult to make something sufficiently general.

Once you're happy with this, I'll delete the old code

### DIFF
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -171,7 +171,10 @@ def parse_parameters(*args):
             raise err
 
     if model['input_output']['format'] is None:
-        model['input_output']['format'] = os.path.splitext(uv_path)[1][1:]
+        path, format = os.path.splitext(uv_path)
+        if format in {'.gz', '.bz2'}:
+            format = os.path.splitext(path)[1]
+        model['input_output']['format'] = format[1:]
 
     param_path = save_prefix + '_frank_used_pars.json'
 
@@ -241,12 +244,9 @@ def alter_data(u, v, vis, weights, model):
 
     wcorr_estimate = None
     if model['modify_data']['correct_weights']:
-        wcorr_estimate, weights = utilities.apply_correction_to_weights(u, v,
-                                                                        vis.real,
-                                                                        weights
-                                                                        )
+        weights = utilities.estimate_weights(u, v, vis, use_median=True)
 
-    return u, v, vis, weights, wcorr_estimate
+    return u, v, vis, weights
 
 
 def determine_geometry(u, v, vis, weights, model):
@@ -534,7 +534,7 @@ def main(*args):
 
     if model['modify_data']['baseline_range'] or \
             model['modify_data']['correct_weights']:
-        u, v, vis, weights, _ = alter_data(
+        u, v, vis, weights = alter_data(
             u, v, vis, weights, model)
 
     geom = determine_geometry(u, v, vis, weights, model)
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -232,9 +232,6 @@ def alter_data(u, v, vis, weights, model):
     -------
     u, v, vis, weights : Parameters as above, with any or all altered according
     to the modification operations specified in model
-    wcorr_estimate : None if model['modify_data']['correct_weights'] = False,
-    else the correction factor by which the weights were adjusted (see
-    frank.utilities.apply_correction_to_weights)
     """
 
     if model['modify_data']['baseline_range']:
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -383,8 +383,16 @@ def estimate_weights(u, v, V, nbins=300, log=True, use_median=False):
     if use_median:
         return np.full(np.len(u), np.median(var[uvBin.bin_counts > 1]))
     else:
-        no_var = uvBin.bin_counts == 1
-        var[no_var] = np.median(var[~no_var])
+        # For bins with 1 uv point, use the average of the adjacent bins
+        no_var = (uvBin.bin_counts == 1).nonzero()[0]
+        ip = no_var+1
+        im = no_var-1
+        while np.any(uvBin.bin_counts[ip] == 0):
+            ip[uvBin.bin_counts[ip]==0] += 1
+        while np.any(uvBin.bin_counts[im] == 0):
+            im[uvBin.bin_counts[im]==0] -=
+        var[no_var] = 0.5*(var[im] + var[ip])
+
 
         bin_id = uvBin.determine_uv_bin(q)
         assert np.all(bin_id != -1), "Error in binning" # Should never occur
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -385,11 +385,10 @@ def estimate_weights(u, v, V, nbins=300, log=True, use_median=False):
     else:
         # For bins with 1 uv point, use the average of the adjacent bins
         no_var = (uvBin.bin_counts == 1).nonzero()[0]
-        ip = no_var+1
-        im = no_var-1
-        while np.any(uvBin.bin_counts[ip] == 0):
+        ip = no_var+1 ; im = no_var-1
+        while np.any(uvBin.bin_counts[ip] == 1):
             ip[uvBin.bin_counts[ip]==0] += 1
-        while np.any(uvBin.bin_counts[im] == 0):
+        while np.any(uvBin.bin_counts[im] == 1):
             im[uvBin.bin_counts[im]==0] -= 1
         var[no_var] = 0.5*(var[im] + var[ip])
 
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -369,9 +369,9 @@ def estimate_weights(u, v, V, nbins=300, log=True, use_median=False):
     q = np.hypot(u,v)
     if log:
         q = np.log(q)
-
-    q -= q.min()
-    bin_width = q.max() / nbins
+        q -= q.min()
+        
+    bin_width = (q.max()-q.min()) / nbins
 
     uvBin = UVDataBinner(q, V, np.ones_like(q), bin_width)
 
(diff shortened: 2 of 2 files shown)


frank/utilities.py
```diff
@@ -115,6 +120,45 @@ def __init__(self, uv, V, weights, bin_width):
...
+        
+        Parameters
+        ----------
+        uv : array, unit = :math:`\lambda`
+            Baselines to determine the bins of
+        
+        Returns
+        -------
+        idx : array,
+            Bins that the uv point belongs to. Will be -1 if the bin does not
```
jeffjennings, line 139: I don't quite follow the rationale for using `-1`. If any of the uv points really do belong to the last bin, won't the return be degenerate with that for points that are at a baseline beyond the last bin? Would it be more intuitive to just return `float('inf')`?

Should there also be a special case (unique return) for points that are at a smaller baseline than the first bin, say `float('-inf')`?

frank/utilities.py
```diff
@@ -272,7 +321,77 @@ def cut_data_by_baseline(u, v, vis, weights, cut_range):
 
     return u_cut, v_cut, vis_cut, weights_cut
 
+def estimate_weights(u, v, V, nbins=300, log=True, use_median=False):
+    r"""
+    Estimate the weights using the variance of the binned visibilities. 
+
+    The estimation is done assuming that the variation in each bin is dominated
```
jeffjennings, line 328: Sorry not to nitpick, just to note that if the bin's sufficiently wide, the variance may also be dominated by real variations in the visibilities for an axisymmetric, deprojected dataset.

frank/utilities.py
```diff
@@ -272,7 +321,77 @@ def cut_data_by_baseline(u, v, vis, weights, cut_range):
...
+    visibilities.
+
+    Parameters
+    ----------
+    u, v : array, unit = :math:`\lambda`
+        u and v coordinates of observations (deprojected).
+    V : array, unit = Jy
+        Obsersed visibility. If complex, the weights will be computed from the
+        average of the variance of the real and imaginary components, as in
+        CASA's statwt. Otherwise the variance of the real part is used.
```
jeffjennings, line 342: Can you add the link to statwt, https://casa.nrao.edu/docs/taskref/statwt-task.html?

frank/fit.py
```diff
@@ -241,12 +244,9 @@ def alter_data(u, v, vis, weights, model):
...
     wcorr_estimate = None
     if model['modify_data']['correct_weights']:
-        wcorr_estimate, weights = utilities.apply_correction_to_weights(u, v,
-                                                                        vis.real,
-                                                                        weights
-                                                                        )
+        weights = utilities.estimate_weights(u, v, vis, use_median=True)
 
-    return u, v, vis, weights, wcorr_estimate
+    return u, v, vis, weights
```
jeffjennings, line 249: Can you remove `wcorr_estimate` from the `Returns` in the docstring for `alter_data`?

frank/utilities.py
```diff
@@ -272,7 +321,77 @@ def cut_data_by_baseline(u, v, vis, weights, cut_range):
...
+    -------
+    weights : array,
+        Esimtate weight for each uv point.
+
+    Notes
+    -----
+        - This function does not support datasets where the weights may be
+          expected to be different for different baselines, e.g. if the data
+          includes visibilities from telescopes of different sizes
+        - Bins with only one uv point will not have a variance estimate. The
```
jeffjennings, line 363: Would it make more sense to use the variance from the 2 adjacent bins instead of the median across all baselines for bins with only 1 uv point?

frank/utilities.py
```diff
@@ -272,7 +321,77 @@ def cut_data_by_baseline(u, v, vis, weights, cut_range):
...
+          median across all baselines is used instead.
+    """
+    logging.info('  Estimating visibility weights.')
+
+    q = np.hypot(u,v)
+    if log:
+        q = np.log(q)
+
+    q -= q.min()
```
jeffjennings, line 372: I don't follow why you need to do this instead of leaving `q` as the absolute baselines?

frank/utilities.py
rbooth200: Idx is an array of int, not float, so something is needed to specify a bad value. -1 is fine for a bad value, but I'm happy with other choices.

frank/utilities.py
rbooth200: Ok, feel free to update

frank/utilities.py
rbooth200: I'd rather not, since that page might change

frank/utilities.py
rbooth200: What if the adjacent bins are empty?

frank/utilities.py
rbooth200: Several reasons: 

1. For log, q.min() could be negative
2. The nbins are more effectively used spreading them from q.min() to q.max() than 0 to q.max(), especially for log.

frank/fit.py
rbooth200: Done

frank/utilities.py
jeffjennings: I thought `float(`inf`) behaves as an integer, e.g., https://stackoverflow.com/questions/40445920/represent-infinity-as-an-integer-in-python-2-7.

If so, I think this would be good to have `float(`-inf`) and `float(`inf`) for baselines smaller (larger) than the left edge of the first bin (right edge of the last bin).

frank/utilities.py
jeffjennings: The nearest non-empty bin on either side? That still seems a bit more accurate to me than the variance across the full baseline range. I realize in many cases the difference between these 2 results can be small, but I can think of edge cases such as a really skewed baseline distribution where the 'nearest neighbor' estimate would be more accurate.

frank/utilities.py
jeffjennings: Done in ff443e5

frank/utilities.py
jeffjennings: `q.min()` could be negative for log if `q.min` is < 1 yes, but you'll never have a baseline at 1 \lambda for mm observations! 

Couldn't you just do `bin_width = (q.max() - q.min()) / nbins` in l.374?

I know this is all internal to `estimate_weights`, so if you're adamant it needs to be done, ok, but I just think it's better practice / more foolproof to keep the baselines in absolute units.

frank/utilities.py
rbooth200: What about ones in between?

frank/utilities.py
rbooth200: You still need to make sure that all q > 0

rbooth200: I could only do it for the log case

jeffjennings: I must be missing something, when will q ever be negative? u and v can individually be negative, but q is always positive, it's a physical distance.

rbooth200: log(q) < 0 if q < 1

frank/utilities.py
jeffjennings: From the link, "You can't represent infinity as an integer, but the floating point representation has a special case to allow it."

Sorry what do you mean by "ones in between"?

frank/utilities.py
jeffjennings: Right, which is why I noted q = 1 is a baseline at 1 \lambda, which you'll never have (can't separate antennas by ~1 mm).

frank/utilities.py
rbooth200: > From the link, "You can't represent infinity as an integer, but the floating point representation has a special case to allow it."

Yes indeed, but we are using integers not floating points for index. floating point indices do not make sense and are forbidden in python.

There are empty bins that have been removed from the UVDataBinner

frank/utilities.py
rbooth200: I've changed the code now. The original worked more like np.histogram with bins=nbins though.

rbooth200: It still makes sense to subtract qmin in the log case because loq(qmin) can be ~ 4-5 while log(qmax) - log(qmin) ~ 1-3. You're just creating a bunch of empty bins for no reason.

frank/utilities.py
jeffjennings: > > From the link, "You can't represent infinity as an integer, but the floating point representation has a special case to allow it."
> 
> Yes indeed, but we are using integers not floating points for index. floating point indices do not make sense and are forbidden in python.

Right, that's kind of the point of using `float(inf)` or more generally something that isn't an integer at which `uv` can't be indexed. I thought it confusing to return `-1` for cases where the baseline doesn't belong to any bin.

> 
> There are empty bins that have been removed from the UVDataBinner

Ok that's what I thought, but then any points "in between" have to belong to a bin, because they've populated it, so that baseline range is no longer empty/removed, right?

frank/utilities.py
rbooth200: Done

frank/utilities.py
jeffjennings: ok yes that's true, I see. Thanks for updating where possible.

frank/utilities.py
rbooth200: The new code can fail if there is only one baseline in the min/max bin

frank/utilities.py
rbooth200: > Right, that's kind of the point of using float(inf) or more generally something that isn't an integer at which uv can't be indexed. I thought it confusing to return -1 for cases where the baseline doesn't belong to any bin.

But it _has_ to be an int because I'm returning an array of indices.

> Ok that's what I thought, but then any points "in between" have to belong to a bin, because they've populated it, so that baseline range is no longer empty/removed, right?

The only time you'll get points in empty bins, before the first, or after the last is if the uv set passed in differs from the set used to create the UVDataBinner. In that case all three are possible. If the datasets are the same, then none of them are possible.

PR #105, opened 2020-04-08
Update `alter_data` call ordering in `fit.py`

### LINKED ISSUE #104: alter_data location
In fit.py alter_data is called before the geometry fit. Is this really desired? Both the uv cut and the weight determination make more sense with the de-projected data.

jeffjennings:
Fixes #104


rbooth200:
This still doesn't use the deprojected data though

jeffjennings:
> This still doesn't use the deprojected data though

You're right; I think it would be much cleaner to do this inside `geometry.py` then (if `alter_data` is called). That prevents having to make separate cases for `FrankFitter` called further down in `fit.py` (because the geometry would already be determined if `alter_data` is called).

rbooth200:
No I disagree, it breaks the "1 function (or class), 1 responsibility rule". The geometry object is responsible only for the geometric corrections.

You should just pass the geometry object into this function, use `apply_correction` to get the corrected visibilities.

rbooth200:
@jeffjennings Are you happy with this implementation? Its the simplest option

jeffjennings:
I'm not crazy about how the geometry is determined with the full baseline distribution rather than the truncated distribution, but it's not the end of the world. I fixed a minor typo and am good to merge!

rbooth200:
That's a fair point, one could always call fit_geometry twice if altering the data? Is that preferred. It is kind of a chicken and egg problem though isn't it? It makes sense to fit the geometry with the truncated data but truncate the deprojected data

jeffjennings:
Yeah you're right, there doesn't seem to be a cleaner way. I guess once the geometry's already been fitted prior to truncation and the effect of that deprojection imprinted on the baseline distribution, it probably doesn't help to re-fit with the truncated data. I think how you have it now works.

rbooth200:
Great, I will merge

PR #114, opened 2020-04-24
Add note to docs on python3 requirement

jeffjennings:
This PR mentions explicitly in the docs main page that frank requires Python >= 3.0.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -1,5 +1,6 @@
 Installation
 ============
+``frank`` requires Python >= 3.0 and does not support compatibility with Python 2.x.
 
 With pip
 --------
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -1,6 +1,6 @@
 Installation
 ============
-``frank`` requires Python >= 3.0 and does not support compatibility with Python 2.x.
+``frank`` requires Python 3 and is tested for Python 3.7. Usage under Python 2.x is not supported.
 
 With pip
 --------
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -33,12 +33,6 @@ interpreting its results, and hosts the code's API.
     Github <https://github.com/discsim/frank>
     Submit an issue <https://github.com/discsim/frank/issues>
 
-Notes
-=====
-``frank`` requires Python >= 3.0 and does not support compatibility with Python 2.x.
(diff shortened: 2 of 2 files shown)


rbooth200:
Can you add the other requirements while your there? Including the ones for the normal code, tests and docs build maybe?

jeffjennings:
> Can you add the other requirements while your there? Including the ones for the normal code, tests and docs build maybe?

The required python packages are forced in `requirements.txt`, where you can also specify a version, e.g., `numpy>=x.x`. 

Is that what you mean?

rbooth200:
If you're going to add the requirements of python3 to the docs, you should probably add the other requirements to the docs at the same time? I thought that was clear enough

jeffjennings:
The package requirements really belong in `requirements.txt`, you just can't force a python version there, or I would've put this there. I've extended the note in `index.rst` though to mention the packages listed in `requirements.txt`. We don't need to list the packages needed to build the docs.

rbooth200:
I'm not asking for changes to requirements. Thanks for adding it to the docs though

mtazzari:
Sorry, I've seen the latest commit only now; I was writing the review and the page didn't update.
FIne like that. Just move it to the Installation page and we close this.

jeffjennings:
Good point, `install.rst` is a better place for this. Done.

jeffjennings:
Good point, `install.rst` is a better place for this. Done.

docs/install.rst
```diff
@@ -1,5 +1,6 @@
 Installation
 ============
+``frank`` requires Python >= 3.0 and does not support compatibility with Python 2.x.
```
mtazzari, line 3: We cannot say that requires Python >=3.0 because it would imply that it works in all Python versions later than 3.0, which span a lot of years back and I am sure it is not the case.
So, I suggest a more specific:
```suggestion
``frank`` requires Python 3 and is tested for Python 3.7. Usage under Python 2.x is not supported.
```
I've opened #116 to address this and discover whether we can support Python 3.6 and 3.8 as well.
I've opened #115 for the analogous with packages.

jeffjennings: Thanks Marco for catching this and for opening the other issues, definitely important to check these things.

PR #117, opened 2020-04-27
Matrix of tests on Circle CI

rbooth200:
This begins a set of tests on circle ci checking the code against different python versions (3.x).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
           command: |
             python3 -m venv venv
             . venv/bin/activate
-            sudo apt-get install libblas3 liblapack3 liblapack-dev libblas-dev
+            sudo apt-get install gcc gfortran python-dev libopenblas-dev liblapack-dev cython
             pip install scipy==<<parameters.scipy-version>>
             pip install -r requirements.txt
             pip install .[test]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,13 +2,16 @@
 #
 # Check https://circleci.com/docs/2.0/language-python/ for more details
 #
-version: 2
+version: 2.1
 
 jobs:
-  test-3.6: &test-template
+  test:
+    parameters:
+      python-version:
+        type: string
     docker:
       # specify the version you desire here
-      - image: circleci/python:3.7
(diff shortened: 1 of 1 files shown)


rbooth200:
When you've got time to look at this please do. We'll need to change the requirements for merge to master when its ready.

mtazzari:
Hi Richard, nicely done. 
1) I see one issue with `ci/circleci:test` that is expected but actually doesn't exist because `test` is expanded into `test-3.6`, `test-3.7`, `test-3.8`. I think this is a bug in github or circleci...Maybe we override the need for that for the merge.
2) what do you mean by "We'll need to change the requirements for merge to master when its ready."? Are you referring to the issue I've just mentioned in (1)?
3) What happened with the lapack, and all these dependencies:
`nstall gcc gfortran python-dev libopenblas-dev liblapack-dev cython`?
Why did you add and them remove them? Just to understand whether they are needed or not..

A part from fixing 1) and 2) is there anything else that needs to be done?

rbooth200:
Hi Marco, Yes your 1) is what I was referring to in my 2). I don't know whether its a bug or expected. I suspect that it might be expected, in which case we can just change the requirements to these three tests?

The lapack stuff was mean trying to use the PR to work out which versions of numpy/scipy we are compatible with so that I could also add those requirements to this PR.  It was pain to get them to install on circleci so I don't yet know the answer, I'm currently fighting with doing it locally. If you're happy with the PR as, feel free to make accept it as, I can open another with the numpy/scipy requirements.

mtazzari:
1) Fine by me requesting the three test to pass, but I wouldn't know how/where to change that. I think GitHub automatically infers them form the circleCI config. Do you know where we can change that?
After this is merged into master, this change would be needed forever. So, if we don't manage to change it we may need to switch off requirement of tests passing before merging, which is a bit annoying...

2) Re scipy/numpy. Sure, we can address it in another PR. 

So, as soon as we are happy with (1), I am happy to accept and merge.

rbooth200:
They can be changed in settings/branches - I'll do it.

rbooth200:
Done - its ready when you are

rbooth200:
@mtazzari I'm happy with this when you are. There are a few things waiting on it

PR #119, opened 2020-04-27
Added core requirements

rbooth200:
I've updated the requirements to be the minimum allowed for numpy, scipy and matplotlib. 

numpy>= 1.11 is required by matplotlib (>=3), but I couldn't get it to install, 1.12 is the lowest I could confirm works.
scipy 0.18 is the first version that provides the correct geometry fit (least_squares did not exist before 0.17, and 0.17 gives a slightly different result).

matplotlib > 3 is required for min_exponent in the axes.formatter rcPararms. There is some weird problem with the text line [here](https://github.com/discsim/frank/blob/262333ab7d438f65970f54c34bf4b84b3b2c79d9/frank/make_figs.py#L142) hence 3.1.

I haven't generated the versions for the tests / docs build.


rbooth200:
The tests and docs build have extra requirements (like pytest and sphinx). I haven't determined the versions needed for those to work.

PR #122, opened 2020-04-30
Exposed sub modules

rbooth200:
Without these lines `import frank` doesn't do anything useful. Since the code is small I don't see a reason not to import the core library at least.

I had to update setup.py to prevent from importing the submodules and hitting problems with the dependencies.


jeffjennings:
This looks good to me, is it ready for a review?

rbooth200:
Yes

PR #128, opened 2020-05-20
Update docs typos, minor rephrasing, fix dead links

jeffjennings:
This PR fixes a few typos in the docs, updates dead hyperlinks in the docs, and fixes the last of the `WARNING`s one gets when building the docs (`make html`).

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -344,9 +344,9 @@ def estimate_weights(u, v, V, nbins=300, log=True, use_median=False):
 
     The estimation is done assuming that the variation in each bin is dominated
     by the noise. This will be true if:
-        1) The source is axi-symmetric,
-        2) The uv-points have been deprojected,
-        3) The bins are not too wide,
+    1) The source is axi-symmetric,
+    2) The uv-points have been deprojected,
+    3) The bins are not too wide,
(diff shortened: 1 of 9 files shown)


frank/utilities.py
```diff
@@ -344,9 +344,9 @@ def estimate_weights(u, v, V, nbins=300, log=True, use_median=False):
 
     The estimation is done assuming that the variation in each bin is dominated
     by the noise. This will be true if:
-        1) The source is axi-symmetric,
-        2) The uv-points have been deprojected,
-        3) The bins are not too wide,
+    1) The source is axi-symmetric,
```
jeffjennings, line 347: I only did this because building the docs complains about `unexpected indentation` here.

rbooth200:
If you've verified that all of the new links work, I'm fine with this

jeffjennings:
thanks for reviewing so fast!  I have meticulously clicked each one.

I guess we decided to get rid of `docs/_build` in the repo, which is why they were all dead (they now all point via the directory structure in the docs).

rbooth200:
docs/_build is a temporary working directory so can't be relied on

jeffjennings:
Yeah all the better that the hyperlinks no longer use `docs/_build`

PR #129, opened 2020-05-21
Add ADS frank methods paper link

### LINKED ISSUE #42: add ADS links
Update README and `index.rst` (including sidebar link to papers citing frank) once we have links.

jeffjennings:
Adds the link to the repo README and the docs' index

Fixes #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,9 @@ The [docs](https://discsim.github.io/frank/) have it all.
 Attribution
 -----------
 If you use **frank** for your research please cite Jennings, Booth, Tazzari et al. 2020 MNRAS (accepted)
-[[MNRAS]](https://academic.oup.com/mnras/advance-article/doi/10.1093/mnras/staa1365/5838058?guestAccessKey=7f163a1f-c12f-4771-8e54-928636794a5b) [[arXiv]](https://arxiv.org/abs/2005.07709)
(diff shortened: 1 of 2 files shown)


rbooth200:
I don't think this is the final ADS link. I can accept the PR, but we shouldn't close the issue.

jeffjennings:
I checked with another paper that the ADS link with the arxiv identifier in the link redirects to the updated ADS link with the jouranl identifier, so I think the current link is persistent.

rbooth200:
Can you find confirmation on ADS that this is always true?

jeffjennings:
This excerpt from http://adsabs.harvard.edu/abs_doc/faq.html is roughly similar.
_Preprints_
_Preprints submitted to the arXiv and which are cited using the proper identifier (e.g. astro-ph/0012345) will be properly assigned to the e-print record in question. If the e-print has been matched to its corresponding published paper, the citation will be "transferred" to the published paper, which in ADS is considered the canonical record for the paper in question._

You can see it works e.g. with this MNRAS paper, 
_https://ui.adsabs.harvard.edu/abs/2018arXiv180300571J/abstract_ redirects to https://ui.adsabs.harvard.edu/abs/2018MNRAS.477.4131J/abstract

rbooth200:
Ok, we can always correct it later if we find a problem.

PR #132, opened 2020-06-08
Uvtable format

jeffjennings:
I noticed the .npz UVTable column format (u, v, V, weights) differs from the .txt column format (u, v, Re, Im, weights). I think it best that the same column format is assumed (for loading and saving a UVTable) across all file extensions. So this PR just makes the five column format (separating V into Re and Im) the default. This is consistent with the documentation and I think the slightly less erorr-prone choice (e.g., a UVTable with the column V that stored the Im(V) component as a number _not_ multiplied by `1j` would currently silently result in an unexpected output from `io.load_uvtable` and `io.save_uvtable`).

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -14,7 +14,8 @@ Perform a fit from the terminal
 To perform a quick fit from the terminal, only a UVTable with the data to
 be fit and a *.json* parameter file (see below) are needed. A UVTable can be extracted
 with CASA as demonstrated in `this tutorial <tutorials/mstable_to_uvtable.rst>`_.
-The column format should be `u [\lambda]     v [\lambda]      Re(V) [Jy]     Im(V) [Jy]     Weight [Jy^-2]`.
+The column format should be either `u [\lambda]     v [\lambda]      Re(V) [Jy]     Im(V) [Jy]     Weight [Jy^-2]`
+or `u [\lambda]     v [\lambda]      V (Re + Im * 1j) [Jy]     Weight [Jy^-2]`.
 
 You can quickly run a fit with the default parameter file, `default_parameters.json` (see below),
 by just passing in the filename of the UVTable to be fit with the `-uv` option. The UVTable can be a `.npz`, `.txt` or `.dat`.
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -204,12 +204,14 @@ def load_data(model):
     weights : array, unit = Jy^-2
         Weights assigned to observed visibilities, of the form
         :math:`1 / \sigma^2`
+    ncols : int
+          Number of columns in the UVTable
     """
 
-    u, v, vis, weights = io.load_uvtable(
+    u, v, vis, weights, ncols = io.load_uvtable(
         model['input_output']['uvtable_filename'])
 
-    return u, v, vis, weights
+    return u, v, vis, weights, ncols
 
 
 def alter_data(u, v, vis, weights, geom, model):
@@ -403,7 +405,7 @@ def perform_fit(u, v, vis, weights, geom, model):
         return [sol, None]
 
 
-def run_multiple_fits(u, v, vis, weights, geom, model):
+def run_multiple_fits(u, v, vis, weights, ncols, geom, model):
     r"""
     Perform and overplot multiple fits to a dataset by varying two of the
     model hyperparameters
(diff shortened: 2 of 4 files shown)


rbooth200:
I think Marco needs to comment on this. Personally, I'm against breaking existing code so I'd suggest load allows there to be either V present (which is complex) or Re and Im which are both real. Can you do that?

jeffjennings:
I think adding the option for what column format the input UVTable is in will just lead to confusion and will result in some frank output UVTables being in different format from others. I'd really prefer to stick to a single format so that everything's self-consistent, with that format being what's specified in the current docstrings.

rbooth200:
I understand your reasoning, but you must not break existing functionality without giving people adequate warning. You have three options: 1) Do nothing, 2) Support both, 3) Support both for a time, with a deprecation warning for a period before removing it. 

Since there is no good argument against 2) and 3) involves more work than 1) or 2) I suggest you do 1) or 2).

rbooth200:
Another reason for supporting u, v, V is that it is the format exported by Marco's uvtable, which we encourage people to use to extract the visibilities.

jeffjennings:
> Another reason for supporting u, v, V is that it is the format exported by Marco's uvtable, which we encourage people to use to extract the visibilities. So I think these changes need rolling back.
> 
Marco's uvtable allows multiple formats, including specifying u, v, V or u, v, ReV, ImV (see the paragraph "From version v0.2.6 it is necessary..." here -- https://github.com/mtazzari/uvplot).

> If you want to add support for reading a .npz with the ReV and ImV fields when V is not present, I think that is reasonable, but I think the default read and write for npz must not be changed.
I've now made it possible to load/save UVTables (under any file extension) with either 4 or 5 columns. I've made the default (in `io.save_uvtable`) 5 columns, though the number of columns to save is inherited from the number of columns in the input UVTable (see `io.load_uvtable`). 

I think this is the best compromise, having the default column format for saved UVTables the same regardless of the table's file extension, but inheriting the column format from the loaded UVTable (if running a fit from the terminal).

rbooth200:
UVtable always uses V when saving in binary, see [here](https://github.com/mtazzari/uvplot/blob/918bf01a1f4313664751eb4a16df130e9b4c4e1d/uvplot/uvtable.py#L276). Our default should be the same. I don't see the need for both save formats.

jeffjennings:
ah my myistake, I didn't notice Marco's default for binary. I've updated the default to `ncols=4` in `io.save_uvtable`. I do think that if there's an option to load/save non-binary UVTables with different column formats, there should be the same option for binary UVTables, as I don't see a reason to favor one column format over the other. Increasing the code flexibility to accomodate multiple formats will be most convenient for users in my view, as some people may have binary UVTables with Re(V) and Im(V) saved as separate variables. I think it would be confusing to allow different formats for non-binary file formats, but not for binary file formats.

rbooth200:
Agreed, but it is true that he always stores the visibilities as V (complex) in npz. The ASCII text was always in ReV and ImV, but there's no such thing as a complex datatype in ASCII anyway.

jeffjennings:
Yeah you're right, it's only that I don't see a reason to not offer the flexibility for all file types, with the default being what Marco uses in uvtable.

rbooth200:
Personally, I still think what we had originally is best.

Doesn't numpy.savetext discard the complex part (https://stackoverflow.com/questions/6494102/how-to-save-and-load-an-array-of-complex-numbers-using-numpy-savetxt)? To get around that we'd need to design our own complex ASCII format. Essentially we already have one so I don't see a reason for having two. So I think that the ASCII format doesn't want to be changed from master.

The default npz should also be the same as master for consistency with uvplot. I'm ok with you adding an option to store ReV and ImV in the npz if you really want, but I don't see a good reason for adding it as an option. If we _do_ at it then the function argument determining the format needs to be more descriptive than ncol. After all, what does ncol=3 or ncol=6 mean? 

Basically, I still think we should roll back all of these changes, and just improve the docs.

jeffjennings:
The point of this PR was that it's not consistent to save an output of the code, a UVTable, in different column formats dependent on the file extension (regardless of what's done in _uvtable_). The ASCII and binary formats are consequently not directly compatible, and as I mention in the PR description, master doesn't catch if the variable V in a binary file is of the form xx + yy rather than xx + yy * 1j. That at minimum should be fixed, but I'll delete this PR.

rbooth200:
> master doesn't catch if the variable V in a binary file is of the form xx + yy rather than xx + yy * 1j. That at minimum should be fixed, but I'll delete this PR.

You could (maybe should) add a check that np.iscomplexobj(V) is True

PR #133, opened 2020-06-15
Clarify UVTable column formats based on file extension

jeffjennings:
Clarify in `quickstart.rst`


rbooth200:
Do you want to add the np.iscomplexobj(V) check that I mentioned in the old PR?

jeffjennings:
Done, I just raise an error rather than trying to correct 'V' because if it's not complex in an input UVTable, it may only have the Re(V) component.

rbooth200:
I'm happy with that. I explicitly changed xx to Re(V) and yy to Im(V) for clarity

PR #134, opened 2020-08-07
Fix some miscellaneous bugs and add minor improvements in response to user feedback

jeffjennings:
From emails in the last couple days, users have noted a couple bugs in the code that arise under certain input parameter choices. This PR fixes these and adds more tests for the associated calls. 

There's also a minor expansion a user requested, to fit for only dRA and dDec (providing inc and PA), implemented here.

Can you check especially carefully the changes in geometry.py in ec88bc6?


frank/geometry.py
```diff
@@ -339,6 +344,9 @@ def fit(self, u, v, V, weights):
             Weights on the visibilities
         """
 
+        if self._inc_pa:
```
rbooth200, line 347: Can we have a check to ensure that we are fitting something? i.e. one of inc_pa and phase_centre is None. Same for FitGeometryFourierBessel

frank/parameter_descriptions.json
```diff
@@ -12,12 +12,14 @@
 
   "modify_data" : {
     "baseline_range"   : "(Deprojected) baseline range outside of which visibilities are truncated, unit=[\\lambda]",
-    "correct_weights"  : "Whether to estimate the data's weights (as they may be unknown for mock data)"
+    "correct_weights"  : "Whether to estimate the data's weights (as they may be unknown for mock data)",
+    "norm_wle"         : "Wavelength (unit=[m]) by which to normalize the (u, v) points. Not needed if the (u, v) points are already in units of [\\lambda]"
```
rbooth200, line 16: Can you be explicit that the wavelength is used to convert the uv points to radians? "Normalize" is a bit confusing

frank/tests.py
```diff
@@ -346,6 +347,10 @@ def test_pipeline_no_phase():
     """Check the full fit pipeline when fitting for the disc's inc, PA"""
     _run_pipeline('gaussian', False)
 
+def test_pipeline_no_inc_no_pa():
```
rbooth200, line 350: I can't see where fit_inc_pa is used in `_run_pipeline`

frank/geometry.py
jeffjennings: Good point, fixed in 1eae50b

frank/parameter_descriptions.json
jeffjennings: yeah, fixed in d489249

frank/tests.py
jeffjennings: good catch! I left out the line. Added in 79a2cab

jeffjennings:
I think this is ready to merge now unless you have other requested changes.

PR #136, opened 2020-08-08
Update docs syntax

jeffjennings:
This PR updates the function calls in the docs to be consistent with current code and makes a minor related change to figure generation (adds `figsize` variable to all figures).


docs/quickstart.rst
```diff
@@ -136,21 +136,20 @@ but we're going to mostly avoid those here,
 just to show how to directly interface with the code's internal classes.
 
 First import some basic stuff from frank and load the data
-(again using the DSHARP observations of AS 209, available as a UVTable
+(again using the DSHARP observations of AS 209, available as a UVTable - with some reasonable time- and channel-averaging -
```
rbooth200, line 139: Please add the details...

jeffjennings: I think the time averaging for the AS 209 UVTable is 10 s (or 30 s?) and the channel averaging 1 channel per spw (spectral window). @mtazzari is that correct? 

(I just wanted to add this detail in case people are comparing directly to the MS Table on the DSHARP website, but don't have to.)

jeffjennings:
> Thanks Jeff. What changed n "fitting_procedure.ipynb"

I know it's annoying isn't it that the diff shows all the raw .ipynb code. I just changed the function calls to have the correct arguments for the current version of the code, and I removed a few imports that are no longer used as a result.

rbooth200:
> I know it's annoying isn't it that the diff shows all the raw .ipynb code. I just changed the function calls to have the correct arguments for the current version of the code, and I removed a few imports that are no longer used as a result.

Fine by me.

docs/quickstart.rst
jeffjennings: ...I found the averaging used (stated in the frank ms). Updated in b49c8b1

jeffjennings:
I think this is ready to merge now unless you have other change requests.

PR #137, opened 2020-08-08
Extend functionality of sweep_profile

jeffjennings:
This PR adds the option for `utilities.sweep_profile` to return an image projected by a supplied `geom`.

TODO: 
[x] Add reprojected image to `make_quick_fig`, `make_full_fig`
[x] Add option to set colorscale to for swept images as either 'power' or 'asinh' 
[x] Add option for whether figures plot in linear or log(baseline)
[x] Update relevant figures in docs


frank/utilities.py
```diff
@@ -458,38 +458,77 @@ def draw_bootstrap_sample(u, v, vis, weights):
...
+        cos_i = np.cos(inc)
+        cos_pa, sin_pa = np.cos(pa), np.sin(pa)
 
     xmax = ymax = r.max()
     dr = np.mean(np.diff(r))
+
+    if project:
+        xmax *= abs(cos_i * cos_pa) + abs(sin_pa)
+        ymax *= abs(cos_i * sin_pa) + abs(cos_pa)
+        xmax = ymax = int(max(xmax, ymax) / dr + 1) * dr
```
rbooth200, line 515: Can you explain the logic here?

frank/utilities.py
```diff
@@ -458,38 +458,77 @@ def draw_bootstrap_sample(u, v, vis, weights):
...
+
+        xp  = (xi + dra) * cos_pa + (yi - ddec) * sin_pa
+        yp  = -(xi + dra) * sin_pa + (yi - ddec) * cos_pa
+        xp /= cos_i
+
+        r1D = np.hypot(xp, yp)
+
+    else:
+        xi, yi = np.meshgrid(x, y)
+        r1D = np.hypot(xi, yi)
```
rbooth200, line 531: I don't think this correctly handles both the project / shift cases independently. (e.g. such as shift with no project)

mtazzari:
I haven't had time to check this.
Is this a re-implementation of what is already in galario?
if yes, why are we keeping doing it?

Marco

—
Dr. Marco Tazzari - mtazzari@ast.cam.ac.uk -
https://www.ast.cam.ac.uk/people/marco.tazzari
Postdoctoral Research Associate - office H32  phone +44 (0)1223 766691
Institute of Astronomy, University of Cambridge, Madingley Road, Cambridge,
CB3 0HA, UK

rbooth200:
I think Jeff wanted to be able phase-shift the image in image-space, which is not implemented in galario. But I agree that it could have been done in galario instead

jeffjennings:
Yes this is just an expansion of the existing function `utilities.sweep_profile` that swept a 1D frank profile over 2\pi. The expansion is to also deproject and phase shift the image. I just use this to make an image of the reeprojected, swept frank fit in a couple of the optional figures output by a fit, so that it is more immediately comparable to a CLEAN model of an observation. I don't mean to be redundant with galario, though galario isn't a dependency and so I didn't want to import a function from it. The syntax in the function also isn't identical to what's in galario, so I think it's ok to leave this as a unique function in frank.

frank/utilities.py
jeffjennings: Yeah, this is just precomputing the x and y bounds of the image to account for the 
'stretching' of the swept profile along each axis due to projection; if we just left `xmax = ymax =r.max()`
and imaged a projected source, under sufficiently high inc and PA the disc would not be fully
contained in the image. The third line here sets `xmax = ymax` so that the image will still be a square
(regardless of the disc geometry).

frank/utilities.py
jeffjennings: Yeah I accounted for a case of reprojection with no phase shift, or reprojection and phase shift,
but not phase shift without reprojection, just because I didn't think it would be useful. But
you're right, it should be included; I fixed in ecedf4e.

rbooth200: Fine, but if you're just correcting for PA and inc, isn't it correct to keep xmax=ymax=rmax? You could try to correct for dRA and dDec too, but perhaps its easiest to use xmax=ymax=rmax as default and allow the caller to specify an alternative, if desired?

jeffjennings: I see now, you're right that is simpler. I've set `xmax=ymax=rmax` always, and added function arguments to optionally specify `xmax` and `ymax`, in c5001fb.

frank/utilities.py
```diff
@@ -458,38 +458,84 @@ def draw_bootstrap_sample(u, v, vis, weights):
...
+        cos_pa, sin_pa = np.cos(pa), np.sin(pa)
+
+    if not xmax and not ymax:
+        xmax = ymax = r.max()
+    elif not xmax:
+        xmax = r.max()
+    elif not ymax:
+        ymax = r.max()
+    else:
+        pass
```
rbooth200, line 520:
```suggestion
    if not xmax:
        xmax = r.max()
    if not ymax:
        ymax = r.max()
```

jeffjennings: In my defense, it was a little late ha. fixed in b028bfc

jeffjennings:
> Can we allow the user to specify dx as well? Then I think this is ready to go

Done in 48e7aca. 

I was going to add swept, reprojected images to a couple of the optional figures produced by a fit, but no problem to do this in another PR if you want to merge now.

rbooth200:
I can wait

jeffjennings:
> I can wait

cool I'll add that shortly.

jeffjennings:
Sorry it took so long to get back to this, but I've now finished it. The checklist at the top of the comments describes what's in it. You reviewed it on August 13, so if it's faster, you can just review changes since. Thanks!

rbooth200 (changes requested):
Looks good, just a few small changes.

frank/make_figs.py
```diff
@@ -300,19 +341,26 @@ def make_full_fig(u, v, vis, weights, sol, bin_widths, alpha, wsmooth,
...
     force_style: bool, default = True
         Whether to use preconfigured matplotlib rcParams in generated figure
     save_prefix : string, default = None
         Prefix for saved figure name. If None, the figure won't be saved
     norm_residuals : bool, default = False
         Whether to normalize the residual visibilities by the data's
         visibility amplitudes
+    stretch: string, default = 'power'
+        Transformation to apply to the colorscale. The default 'power' is a
+        power law stretch; the other option is 'asinh', an arcsinh stretch.
```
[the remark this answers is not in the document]
jeffjennings: Good point! Added in 4504b46

rbooth200 (approved):

PR #138, opened 2020-08-11
[Reopen] Fitting geometry when inc, PA are provided

jeffjennings:
Sorry about this, there was a bug I missed in #134, corrected here. In `FitGeometryGaussian`, a user-provided inc and PA were being converted from [rad]--> [deg], even though they were provided in [deg]. I missed this because I was testing with a provided inc and PA of 0. 

I've also added tests for when the inc and PA or phase offset are provided (using nonzero values for the tests).

Because the last few PRs have addressed bugs, I'd also like to push to PyPI after this PR is merged so that `pip install frank` includes these fixes. PyPI would go from `v1.0.1` --> `v1.0.2`.


frank/geometry.py
```diff
@@ -488,7 +488,10 @@ def _gauss_jac(params):
     if phase_centre is not None:
         dRA, dDec = phase_centre
 
-    geometry = inc / deg_to_rad, PA / deg_to_rad, dRA, dDec
+    if inc_pa is None:
+        geometry = inc / deg_to_rad, PA / deg_to_rad, dRA, dDec
+    else:
+        geometry = inc, PA, dRA, dDec
```
rbooth200, line 494: I don't understand why the units are different in these cases. In both cases the units must be radians in the _gauss_fun etc.

jeffjennings: Ok so the issue was that `guess` as initialized in `__init__` didn't change if `inc_pa=True` or `phase_centre=True`, which didn't make sense, because it meant that if one of these two was `True`, the other was being fit for without taking into account that the first was known/given. The `guess` values now update if `inc_pa=True` or `phase_centre=True`. 

I also converted `self._inc_pa`from [deg] --> [rad] in `__init__` so that the units of inc and PA are now [rad] throughout, regardless of whether their values are given.

rbooth200: Thanks, but I think it is better if _fit_geom_gauss uses the same units for its arguments and what it returns. Can you do the conversion inside fit_geom_gauss? The same should be true of the guess parameter

jeffjennings: Ok I've moved all unit conversion/reconversion into `_fit_geometry_gaussian` in b8a6942, and in that commit I've also updated the `guess` if one isn't provided, such that it has inc, PA = 10 deg rather than 0.1 deg. I also reordered `guess` and the `params` order in `_gauss_fun` and `_gauss_jac` so that it's inc, PA, dRA, dDec (consistent with the order of a user-provided guess to avoid confusion).

Separately, in e3f6434, I haven't seen this before, but the tests are passing (on my machine and) in the python 3.7 and 3.8 versions, but not 3.6. This is seemingly due to a (large) floating point error; do you see what's wrong?

jeffjennings: (it's hacky, but we could just reduce the tolerance in the `FitGeometryFourierBessel` test functions from 1e-5 to 1e-4.)

rbooth200: What have you changed that is affecting this?

jeffjennings: I changed the asserted values in the test for `FourierBesselFitter` based on locally running the test. I rechecked this and still get values discrepant with those I reverted back to in the test in 3750964, but not clear to me what's different in my environment that would cause this, or why it passed on CircleCI in python 3.7 and 3.8 but not 3.6; some constant or something must be different between versions.

Anyway after reverting back to the previous asserted values, it passes all tests now.

rbooth200: Did you alter the initial guesses?

jeffjennings: In `tests.FourierBesselFitter` I didn't. In `geometry.FitGeometryGaussian` 

I changed the initial guesses for inc and PA from 0.1 --> 10.0 deg, as 0.1 was I presume meant to be in [rad]. 

I didn't change the initial guesses in `geometry.FitGeometryFourierBessel` (which were already 10 deg for inc and PA).

rbooth200: Ok. I'll see if I can spot something that might have broken it, but it could simply be different versions of the library giving slightly different results.

jeffjennings: yeah I think you're right

PR #139, opened 2020-08-14
Allow UVDataBinner to intake and bin only real component of vis

jeffjennings:
This PR makes `UVDataBinner` consistent with `utilities.estimate_weights`, which allows the option for `V` to only be real. Because `estimate_weights` calls `UVDataBinner`, the latter should then be able to handle a supplied `V` with only real components.

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -56,7 +56,8 @@ class UVDataBinner(object):
     uv : array, unit = :math:`\lambda`
         Baselines of the data to bin
     V : array, unit = Jy
-        Complex visibilities
+        Observed visibility. If complex, both the real and imaginary
+        components will be binned. Else only the real part will be binned.
     weights : array, unit = Jy^-2
         Weights on the visibility points
     bin_width : float, unit = :math:`\lambda`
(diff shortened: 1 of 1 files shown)


jeffjennings:
If someone's generating mock data in 1D (in baseline) and computing the weights with `utilities.estimate_weights`, it could be more convenient for `estimate_weights` to intake baselines rather than `u` and `v` separately. Would you think it reasonable to update the arguments for `estimate_weights` to intake baselines?

jeffjennings:
> def f(u, v, V=None):
>     if V is None:
>         q = np.abs(u)
>         V = v
>     else:
>         q = np.hypot(u,v)

That is clever, but yeah maybe not the cleanest. I've added an optional argument `q` as another approach.

rbooth200:
I don't like this solution because u and V still need to be passed in

jeffjennings:
> I don't like this solution because u and V still need to be passed in

ok, I didn't come up with a good solution that wouldn't change the argument order in the function, so updated to use your hack. The function docstring for `V` explains how to pass in `q`, and a logging message prints when that's done, so it should be pretty foolproof.

rbooth200:
I made v an optional argument so that esimate_weights (u,V=V) will work. If also added example function calls to make it clear how to use this function correctly. For what it's worth, numpy uses this idea:
https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.random.randint.html

jeffjennings:
Cool, I just deleted a couple comments, and it all looks good to me.

PR #144, opened 2020-12-11
fit.py: include call to alter_data if norm_wle is provided

jeffjennings:
`fit.py` wasn't checking whether `norm_wle` was provided in the parameter file in the conditional for whether to call `alter_data`. And it wouldn't be obvious to a user from the logging that this correction hasn't been applied.


jeffjennings:
Hey @rbooth200 sorry to bother, but can you review this when you get a sec (somewhat urgent)? It's just 1 line.

rbooth200 (approved):

PR #149, opened 2021-01-22
quick_fig: ensure frank fit plots in front of data

jeffjennings:
Very minor change in `make_quick_fig` to ensure fit line plots in front of binned visibilities

### DIFF
--- a/frank/make_figs.py
+++ b/frank/make_figs.py
@@ -176,7 +176,7 @@ def make_deprojection_fig(u, v, vis, weights, geom, bin_widths, force_style=True
 
     return fig, axes
 
-  
+
 def make_quick_fig(u, v, vis, weights, sol, bin_widths, dist=None, logx=True,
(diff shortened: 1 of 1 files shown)


rbooth200:
Can you upload a before / after plot?

jeffjennings:
Yeah sure, before (current `master`) and after (this branch):
SR4_continuum_frank_fit_quick_after
SR4_continuum_frank_fit_quick_before

rbooth200:
Does it make sense to have the binned data at the front?

jeffjennings:
I don't think so, in that case you can't see the fit line clearly, even with fairly large bins. If the fit line is instead in front, it's still clear where the data are, as their marker styles are '+' and 'x'.

rbooth200:
I guess I was just thinking the 50 k\lambda bins. But anyway, I do not have a strong opinion either way.

rbooth200 (approved):

PR #150, opened 2021-01-22
estimate_weights flexibility

jeffjennings:
This PR:

- Adds a parameter to the `default_parameters.json` to set the `use_median` flag in the call to `estimate_weights` in `fit.py`

- Sets the default in the `.json` to `false`, i.e., don't set all estimated weights as the single median variance across bins. I've found using the alternative, the baseline-dependent variance, to give a more sensible result for both the binned visibilities (by comparing 1 and 50 k\lambda binned Re(V)) and the corresponding frank fit for some datasets.

- Changes `np.median` to `np.ma.median` in `estimate_weights` to avoid a `MaskedArray` warning:
      ```
      python3.7/site-packages/numpy/core/fromnumeric.py:746: UserWarning: Warning: 'partition' will ignore the 'mask' of the MaskedArray.
  a.partition(kth, axis=axis, kind=kind, order=order)
      ```
I've verified this doesn't change the calculated median.

### DIFF
--- a/frank/default_parameters.json
+++ b/frank/default_parameters.json
@@ -13,6 +13,7 @@
   "modify_data" : {
     "baseline_range"   : null,
     "correct_weights"  : false,
+    "use_median_weight": false,
     "norm_wle"         : null
   },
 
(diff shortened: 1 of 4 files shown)


rbooth200 (approved):
Fine, I prefer the by-the-bin approach anyway

PR #153, opened 2021-02-17
more verbose _HankelRegressor overflow error message

jeffjennings:
A few people now have run into this error message when calling `FrankFitter`,

```
RuntimeWarning: overflow encountered in exp
  pi_new = np.exp(sparse_solve(Tij_pI, beta + np.log(pi)))
ValueError: Bad value in power spectrum. The power spectrum must be postive and not contain any NaN values
```

In all cases (including my own experience), this has been due to one of: 
- the UVtable passing in `u` and `v` in [m] rather than [\lambda],
- the UVtable weights being incorrect (due to some aspect of extracting the UVtable from CASA), 
- the geometry/phase center used for the deprojection being pretty wrong, 
- or more rarely the fit's `rout` and/or `n` needing to be increased.

This PR just makes the above error message more verbose to help point people to these likely underlying causes.

### DIFF
--- a/frank/radial_fitters.py
+++ b/frank/radial_fitters.py
@@ -121,7 +121,12 @@ class _HankelRegressor(object):
             if np.any(p <= 0) or np.any(np.isnan(p)):
                 raise ValueError("Bad value in power spectrum. The power"
                                  " spectrum must be postive and not contain"
-                                 " any NaN values")
+                                 " any NaN values. This is likely due to"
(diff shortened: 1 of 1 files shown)


rbooth200 (approved):
Yeah, recently I saw this with negative weights (which is nonsensical). Early on I would get this error for numerical reasons but I agree that it generally means bad inputs.

jeffjennings:
Cool, thanks for the super fast review!

PR #160, opened 2021-03-17
Log-normal fits

### LINKED ISSUE #165: storing priors in 'sol'
Hey @rbooth200 I think I remember we talked about this before, but could we could store the 5 hyperparameters in `sol` as a dictionary (say, `priors` or similar)? When loading a `sol.obj` saved from a fit, the filename currently includes alpha and wsmooth, but it would be helpful if those were stored in the object so the fit would be more easily reproducible.

### LINKED ISSUE #169: improve clarity on optically thick flux rescaling
Per Marco's suggestion, make it explicit in the docs that we're applying a flux rescaling (leaving this as the default behavior to avoid user confusion when reproducing previous frank fits). Also separate the flux rescaling into its own method in `geometry`, and make the rescaling optional in `FourierBesselFitter`.

rbooth200:
This branch includes my implementation of the log-normal model for brightness reconstruction. This is still single-frequency code. The methods are now fairly mature (they don't crash any more), but I don't think they should be included in the release that is currently in the pipeline.

The code also includes a considerable amount of refactoring. Some of this was inevitable since we now need to support both Gaussian and Log-Normal models. Some of it was in preparation for the multi-frequency code. 

(Jeff): 
- The PR also adds the `info` property to `FrankRadialFit` (with `info` returned by `FrankFitter`), which Fixes #165.
- The PR also adds the option to not rescale the total flux according to the source geometry, which Fixes #169.
- This PR updates the code version from 1.1.0 to 1.2.0; after merge, I'll tag the release on github and update the package on PyPI.


rbooth200:
The tests are failing due to timeout - hopefully this should be fixed when we transition to github actions

jeffjennings:
> This branch includes my implementation of the log-normal model for brightness reconstruction. This is still single-frequency code. The methods are now fairly mature (they don't crash any more), but I don't think they should be included in the release that is currently in the pipeline.

This is all awesome. I'm just thinking it could also be good to wait to merge this until:
- we all discuss the changes soon via zoom, so that Marco and I have a clear understanding of the refactoring 
- Marco and I both review this PR, so that we just get more eyes on these big changes (I don't doubt they're correct, just to get everyone's perspective) 
-  the 'front end' of the code (`fit.py`, the `.json`s) has been updated to support running fits for the log-normal model
 
It would also be useful to discuss how the multi-frequency code will build on these changes.
> 
> The code also includes a considerable amount of refactoring. Some of this was inevitable since we now need to support both Gaussian and Log-Normal models. Some of it was in preparation for the multi-frequency code.

rbooth200:
Thanks Jeff, I'm happy to discuss this on zoom if its the best way forward. I certainly agree that these large changes aren't to be merged to master until we are all happy with them. Feel free to make a to-do list to add support for the fitting script etc.

jeffjennings:
Cool! It's mostly just so that I understand the new/expanded structure more than anything. I figure that would be easiest by zoom as you can walk us through it, and that would be a good opportunity to understand how best the front end should be modified to complement the refactoring (I see it's still just calling `FrankFitter`, so I'm sure it won't be much to change).

jeffjennings:
Hey @rbooth200, are you ok with #25df8bc? I did this to keep `sol.mean` as a valid call to both the Normal and LogNormal models, since that's been the default call up to this branch. Otherwise, everywhere in the code that calls `sol.mean` could be changed to `sol.MAP` in order to work for both the Normal and LogNormal models, but I figured it's the same either way, so better to keep the current behavior.

rbooth200:
Is it so much to change mean to MAP? I suspect not, and the changes would be limited to a couple of files. 

 I'm not keen on this change because the 'best fit' value for the log normal model is not the mean anymore, but the maximum a posteriori.

jeffjennings:
> Is it so much to change mean to MAP? I suspect not, and the changes would be limited to a couple of files.
> 
> I'm not keen on this change because the 'best fit' value for the log normal model is not the mean anymore, but the maximum a posteriori.

ok I undid this. I think it would be helpful to have a property `I` in `FrankRadialFit` to return the MAP for either fit method.

rbooth200:
Yeah, I was using MAP because MAP=mean for the Gaussian - but you could add/use I instead

jeffjennings:
cool I added I (just points to MAP)

rbooth200:
Great, you might want to edit the plotting scripts to work with MAP or I

jeffjennings:
yeah for sure I'll update all of that and the docs soon. It would be cool to have a notebook tutorial in the docs, comparing the normal and lognormal fits to discuss how/why they're different. I can make an example comparison for a disc if you want to write some discussion (will upload here when I can).

rbooth200:
Yeah a tutorial / guide is a good idea. Probably both a tutorial demonstrating how to use them and a guide/FAQ discussing some of the difficulties since LN model isn't as robust

jeffjennings:
Cool I can draft a tutorial demonstrating how to use them. Do you want to draft the guide/FAQ? 

Could you also update `py_API.rst` in the docs to include the new classes?

rbooth200:
> Could you also update py_API.rst in the docs to include the new fitting classes?

Done. I added the essentials, but left out some of the internal routines

rbooth200:
@jeffjennings Can you fix the conflicts here? It should be easy enough - you are more familiar with `make_figs.py` than I.

jeffjennings:
k fixed. I guess it's just the docs/tutorials that need to be updated now for this to be merged (that could be a separate PR though if you'd prefer), or is there anything else you can think of?

rbooth200:
I'm happy once the docs are done. But since this is a big change, I think Marco should look through too.

rbooth200:
Looks good, but maybe you should add a default option of either `None` or `{}` to `FrankGaussianFit` and `FrankLogNormalFit`.

jeffjennings:
cool, ee2e0219 does that

rbooth200:
Great, we just need to do the docs now?

jeffjennings:
yeah, I'll slack you

jeffjennings:
I think 1d70e02 correctly addresses your comment on #167, "You need to update the .predict methods of FrankRadialFit appropriately", but let me know if not.

jeffjennings:
One thing I forgot; these changes definitely warrant a new code version and release. 326e1e2 updates the code version from 1.1.0 to 1.2.0 and updates the changelog in `HISTORY.rst`. Feel free to edit that. 

After this is merged (the tutorial/docs are all I can think of that are left to do), I'll tag master as v.1.2.0 and update the pypi release.

rbooth200:
I've now applied the fix for #172 to this branch. Are we now ready for a merge?

jeffjennings:
Yeah if you'd like; we could open a separate branch for the tutorial/to update the docs. Merging this will update the codebase to v.1.2.0 (following 326e1e2).

rbooth200:
Let's do the docs first - I don't think we should have release with undocumented features.

jeffjennings:
Makes sense. I know you added a few classes to the API, and the tutorial we discussed was one that uses a mock disc with a shelf at the disc outer edge and a gaussian inside a narrow, deep gap (such that the normal model produces a negative flux). Both these features would be at ~RMS noise level in the mock image, and the tutorial would compare the normal and lognormal fits, including some discussion on the limitations/considerations with the lognormal model. You suggested also showing an example of what happens when you use R_out that is too small since it's more of a problem for the lognormal model.

If that still sounds good, I can split the work for a jupyter notebook for this however you'd like.

jeffjennings:
Hey @rbooth200, can you update `HISTORY.rst` with additional changes that are now here?

rbooth200:
I've updaed the history and API docs, plus a few small changes needed to get the docs to compile. From my side, I guess there is still the question of what changes to make to the docs to cover the new features

jeffjennings:
> I've updaed the history and API docs, plus a few small changes needed to get the docs to compile. From my side, I guess there is still the question of what changes to make to the docs to cover the new features

This is what I can think of: 
1. I'll add a note in the docs on the optically thick assumption. 
2. I'll make the tutorial we discussed for running a log-normal fit. 
3. How accessible do you want to make `FrankDebrisFitter`? If it's something that could be useful to more people, we could add a tutorial, and if so:
- do you have a testing script/notebook with an example case that we could turn into the tutorial? 
- The debris fitter could even be run from the command line using a parameter in the `.json` file like "function":{"arguments":"a,b,c","body":"return a*b+c;". 

Other than that:
- do you want to add tests for `FrankDebrisFitter` and `FourierBesselDebrisFitter`? 
- if you could review `ci_github_actions` before we merge this branch, we can merge master into here and the tests will run.

rbooth200:
I like your suggestions for 3, but I think we should leave these until after Seba's paper is done. I have some materials that we could use to get us started, certainly.

frank/fit.py
```diff
@@ -152,6 +152,11 @@ def parse_parameters(*args):
         err = ValueError("method should be 'Normal' or 'LogNormal'")
         raise err
 
+    if model['hyperparameters']['nonnegative'] and model['hyperparameters']['method'] != 'Normal':
+        err = ValueError("nonnegative should only be 'true' if method is 'Normal'")
+        raise err
+
+
```
rbooth200, line 159: I just noticed this - is non-negative = True really an error for the log normal model? Conceptually it's not an error

docs/tutorials/fitting_procedure.ipynb
```diff
@@ -109,6 +109,8 @@
     "\n",
     "You can also supply an initial guess for the source geometry; if running frank from the terminal, setting `geometry : initial_guess` to `true` in your parameter file will use the `inc`, `pa`, and optionally `dra` and `ddec` in that file as the initial guess for the geometry solver. \n",
     "\n",
+    "**By default, the visibility deprojection in frank scales the total flux to account for the difference between the inclined (observed) brightness and the assumed face-on brightness. This reduces the real component of the observed visibility amplitudes by a factor** $1 / \\cos(i)$ **and increases their weights by a factor** $\\cos(i)^2$. The scaling assumes the emission is optically thick, with an integrated 2D flux of $F = \\cos(i) \\int_r^{r=R}{I(r) 2 \\pi r dr}$. To choose not to rescale the visibilities, set `geometry : rescale_flux` to `false` in your parameter file.  \n",
```
rbooth200, line 112: Do you mind if we re-phrase this a bit? It's not wrong, but I find it confusing. Is the following ok?

> **By default, the visibility deprojection in frank scales the total flux assuming that the disk is optically thick everywhere. Since an inclined optically thick disk covers less area on the sky, the total brightness should be lower by a factor **$\\cos(i)$**. Internally, this is done by scaling the observed visibility amplitudes by a factor **$1 / \\cos(i)$** and their weights by **$\\cos(i)^2$**. When using frank to compute the visibilites for the model (the 'predicted visibilities') this scaling will automatically be applied to the model visibilties so that their amplitude matches the observed visibilities. This option is the default because it produces an intensity profile that is consistent with one extracted directly from a 2D image (such as an azimuthal average or  major- or minor- axis cut).  If instead you wish to not rescale the visibilities, as appropriate for an optically thin disk,  set `geometry : rescale_flux` to `false` in your parameter file. \n",

I feel that there might be some argument for being even more explicit by including the solution to the radiative transfer equation for a razor thin disk, $I(r) = \\exp(- \\tau(r)/\\cos i)$, and discussing the limits.

frank/fit.py
jeffjennings: Sorry, I've just come back to this branch and seen these comments. You're right, I've updated in 2d610e3

docs/tutorials/fitting_procedure.ipynb
jeffjennings: Great, I changed the text to your version and added a couple sentences about the solution for a flat disk and its limits in 45dff44.

jeffjennings:
Hey @rbooth200, I'm working now on a jupyter notebook tutorial for the lognormal model, but in it I'd like to use some routines that are on other branches. Given how big this PR already is, I think it'd be best to do a final review of what's here now, merge it, and tag the release as v.1.2.0. Then have a separate branch for the tutorial, and deal with it along with all the other smaller PRs. At that point the release could be tagged as v.1.2.1 and updated on pypi. Does that sound good?

rbooth200:
Yes, fine by me.

rbooth200:
Feel free to merge when you are ready

jeffjennings:
Great, do you mind reviewing the last 5 commits specifically?

jeffjennings (approved):

rbooth200:
I checked over them, and they are fine. A tiny thing we could think about is making the default value of p_0 be None in the json. The frank fitter will choose a sensible value in that case.

jeffjennings:
Good idea, I'll fold that into another PR that's dealing with p_0, #182.

rbooth200:
Ok, great. Merging.

PR #166, opened 2021-09-04
correct colorscale for 2d sweeps in figures

jeffjennings:
Fixes minor bug that applied wrong scaling to colorbars in `make_quick_fig`, `make_full_fig`


rbooth200:
I don't understand where the 10^10 comes from. Why don't vmax/vmin have the same units as the intensity in the call to plot_2d_sweep?

jeffjennings:
good point, 10^{10} was just to normalize the image brightness to 10^{10} jy / sr; #75d8851 updates to pass [jy / sr] into `plot_2dsweep` and optionally rescale the brightness there

rbooth200 (approved):

PR #167, opened 2021-09-09
total flux rescaling 

jeffjennings:
Per Marco's suggestion, this PR makes it explicit in the docs that we're applying a flux rescaling (I've left this the default behavior to avoid user confusion when reproducing previous frank fits). The PR also separates the flux rescaling into its own method in `geometry`, and makes the rescaling optional in `FourierBesselFitter`.

### DIFF
--- a/frank/default_parameters.json
+++ b/frank/default_parameters.json
@@ -25,7 +25,8 @@
     "inc"              : 0.0,
     "pa"               : 0.0,
     "dra"              : 0.0,
-    "ddec"             : 0.0
+    "ddec"             : 0.0,
+    "rescale_flux"     : true
   },
 
   "hyperparameters" : {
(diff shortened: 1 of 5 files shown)


rbooth200:
Maybe you're not finished here but I have a couple of suggestions:
1. Should the argument be called "assume_optically_thick" instead? Since the flux rescaling applied does precisely that. The alternative (not rescaling) is appropriate for the optically thin limit
2. You need to update the .predict methods of `FrankRadialFit` appropriately.

jeffjennings:
good point, c620382 addresses your (1); for your (2) I think it will make sense to merge the lognormal PR, pull those changes into this branch, and update here as I'm not sure how many additional spots in the lognormal codebase will need to be update with the stuff in this branch.

rbooth200:
It might be best to re-do this work on the LN branch and delete this one.

jeffjennings:
ok, I only didn't do it there because that PR already has a lot, but I'll do that and delete this shortly.

rbooth200:
I'd agree with that idea most times

PR #170, opened 2021-09-22
Plots cleanup

jeffjennings:
This PR just combines 2 functions in `plot.py` and cleans up a couple minor things in `make_figs.py`.

### DIFF
--- a/frank/make_figs.py
+++ b/frank/make_figs.py
@@ -329,7 +329,9 @@ def make_quick_fig(u, v, vis, weights, sol, bin_widths, priors, dist=None,
             ax2.set_xscale('log')
             ax3.set_xscale('log')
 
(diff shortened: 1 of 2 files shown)


jeffjennings:
I'd like to further expand and generally improve the plots/figure pipeline, but it's useful to merge this now before it becomes too big/generic.

frank/constants.py
```diff
@@ -22,5 +22,5 @@
 from scipy.constants import c
 
 rad_to_arcsec = 3600 * 180 / np.pi
-sterad_to_arsec = rad_to_arcsec ** 2
+sterad_to_arcsec = rad_to_arcsec ** 2
```
rbooth200, line 25: Just a small comment - this is only valid for small angles of course..

PR #173, opened 2021-11-29
Fix for predicting visibilities with a different geometry

### LINKED ISSUE #172: Predict using a different geometry
Hi! I think that there is a small error when trying to predict the visibilities using a different geometry.

Right now the code is:
```
 if geometry is not None:
        u, v = self._geometry.deproject(u, v)
```
but I think that the correct one should be:

```
 if geometry is not None:
        u, v = geometry.deproject(u, v)

```

Thanks!!

rbooth200:
Fixes #172 and adds a regression test.

This needs to be propagated into the log normal model etc, which will be done after this PR is merged

### DIFF
--- a/frank/radial_fitters.py
+++ b/frank/radial_fitters.py
@@ -320,7 +320,7 @@ class _HankelRegressor(object):
             geometry = self._geometry
 
         if geometry is not None:
-            u, v = self._geometry.deproject(u, v)
+            u, v = geometry.deproject(u, v)
 
         q = np.hypot(u, v)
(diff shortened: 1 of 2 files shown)


jeffjennings (approved):

jeffjennings:
merge when you're ready! might be helpful to make another issue for this that will get closed by the `lognormal` branch merge, just to remember

rbooth200:
Thanks - rather than doing another issue I'll fix it immediately

PR #178, opened 2022-11-03
Python versions

jeffjennings:
- Adds Python 3.10, 3.11 (most recent versions) to tests
- Removes support for python 3.6 (it's several months past end-of-life -- https://www.python.org/dev/peps/pep-0440/#compatible-release)
- Updates minimum scipy version (it was only 0.18, the most recent version is 1.9)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
 
     steps:
       - name: Checkout code
@@ -31,7 +31,7 @@ jobs:
 
       - name: Install
         run: pip install .
-        
+
       - name: Run unit tests
         run: |
           mkdir -p test-reports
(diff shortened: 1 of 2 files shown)


rbooth200:
Before I review: what is the logic for forcing uses to have python 3.7 and greater? I appreciate we should no longer test it on python 3.6 due EOL, but if the code works why stop people?

jeffjennings:
I see what you mean, though I think it makes sense to keep the required python versions current with the versions that python itself supports, otherwise there could be an edge case where a part of the code works with the current python version but not the oldest one we support. While the tests would catch that, we'd have to keep 2 versions of a given routine. 

Also if we keep adding tests for each new python version as I think we should, at some point it will become kind of crazy to be testing on n versions of python, rather than just the ones that are supported by python (the most recent 5 versions).

rbooth200:
Ok - well my view is that we should keep testing 3.6 until either, 1) it stops working properly or 2) github stops providing a 3.6 platform to test it on. There are probably lots of people still using python 3.6 despite EOL status (e.g. it ships with ubuntu 18.04 which is supported until April 2023). It doesn't seem in our interest to not allow them to use the code if it does not cost us anything to do so.

jeffjennings:
Ok that's fine. I think eventually we should stop supporting sufficiently old versions, but I've added 3.6 back to the tests and minimum requirement in afad00e.

rbooth200:
Agreed.

rbooth200 (approved):

PR #179, opened 2022-11-03
Convenience functions

jeffjennings:
Adds a few very basic convenience functions


frank/utilities.py
```diff
@@ -25,6 +25,27 @@
 
 from frank.constants import deg_to_rad, sterad_to_arcsec
 
+def baselines(u, v):
```
rbooth200, line 28: Should baselines (optionally) take a geometry object and deproject for you? These are convenience functions after all, and that would be convenient..

jeffjennings: Yeah I have that functionality in #183, but want to finish and talk about that today before requesting review on it. I just wanted a function here to strictly find the hypotenuse, as I pictured a deprojection routine as returning u and v separately. I could change the name here to `uvdist` or something though.

rbooth200: Well ok - but do we really need a function that is just another name for numpy.hypot?

jeffjennings: I thought it might be useful for someone who needs the conversion to be explicit (a student or something), but it's not critical of course. Removed in c49517c.

rbooth200 (approved):

PR #181, opened 2022-11-06
normalize_uv generalization 

jeffjennings:
Generalizes `utilities/normalize_uv` to handle pointwise wavelengths

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -334,8 +334,9 @@ def normalize_uv(u, v, wle):
     ----------
     u, v : array, unit = [m]
         u and v coordinates of observations
-    wle : float, unit = [m]
-        Observing wavelength of observations
+    wle : float or array, unit = [m]
+        Observing wavelength of observations. If an array, it should be the
+        pointwise wavelength for each (u,v) point
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


frank/utilities.py
```diff
@@ -347,6 +348,11 @@ def normalize_uv(u, v, wle):
     logging.info('  Normalizing u and v coordinates by provided'
                  ' observing wavelength of {} m'.format(wle))
 
+    if type(wle) is not float or int:
+        if len(wle) != len(u):
+            raise ValueError("len(wle) = {}. It should be"
+                             "equal to len(u) = {}".format(len(wle), len(u)))
+
```
rbooth200, line 355:
```suggestion
    wle = np.array(wle, dtype='f'8)
    if  len(wle) != 1 and len(wle) != len(u):
        raise ValueError("len(wle) = {}. It should be equal to "
                         "len(u) = {} (or 1 if all wavelengths are the same)".format(len(wle), len(u)))
```
This version handles lists properly, and numpy will do the type checking.

rbooth200 (changes requested):

frank/utilities.py
jeffjennings: Perfect, thank you.

frank/utilities.py
```diff
@@ -347,6 +348,10 @@ def normalize_uv(u, v, wle):
     logging.info('  Normalizing u and v coordinates by provided'
                  ' observing wavelength of {} m'.format(wle))
 
+    wle = np.array(wle, dtype='f8')
```
rbooth200, line 351:
```suggestion
    wle = np.atleast_1d(wle, dtype='f8')
```

rbooth200 (commented):
I checked, array should be atleast_1d..

rbooth200 (approved):

PR #182, opened 2022-11-08
Pass p_0 into FrankFitter in fit.py

jeffjennings:
Adds `p_0` to the call to `FrankFitter` in `fit.py` (the value specified in the parameter file was being silently ignored).

Also changes default p0 in the parameter file to 'null'.


rbooth200 (approved):
Looks fine to me

PR #183, opened 2022-11-08
Quick fit access 

jeffjennings:
Adds a couple very simple functions to use a `sol` object to deproject a visibility distribution and to get the visibility fit. They're ~one-liners, but useful for interacting with a `sol` object (e.g. I often end up doing theses things when interacting with a fit).


frank/utilities.py
```diff
@@ -729,3 +729,82 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
...
+        Complex, deprojected visibility amplitudes
+
+    """
+
+    geom = sol.geometry
+
+    # deproject and phase center the visibilities
+    ud, vd, Vd = geom.apply_correction(u, v, V)
+
+    return ud, vd, Vd
```
jeffjennings, line 763: Should this also intake, scale and return the weights?

PR #184, opened 2022-11-08
Generic DHT convenience function

jeffjennings:


frank/utilities.py
```diff
@@ -729,3 +729,63 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
     I_smooth /= counts
 
     return I_smooth
+
+
+def generic_dht(x, f, Rmax=2.0, N=1000, direction='forward', inc=None):
```
jeffjennings, line 734: This should also be able to return the transform at specified points (not just the collocation points).

frank/utilities.py
```diff
@@ -851,3 +853,72 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
...
+    DHT = DiscreteHankelTransform(Rmax=Rmax / rad_to_arcsec, N=N, nu=0)
+    geom = FixedGeometry(inc, 0, 0, 0)
+    VM = VisibilityMapping(DHT, geom)
+
+    unit_convert = False
+
+    if direction == 'forward':
+        # map the profile f(x) onto the DHT collocation points.
+        # this is necessary for an accurate transform.
+        y = np.interp(DHT.r * rad_to_arcsec, x, f)
```
rbooth200, line 904:
```suggestion
        y = np.interp(VM.r, x, f)
```

frank/utilities.py
```diff
@@ -851,3 +853,72 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
...
+
+        if grid is None:
+            grid = DHT.q
```
rbooth200, line 907:
```suggestion
            grid = VM.q
```

frank/utilities.py
```diff
@@ -851,3 +853,72 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
...
+
+        # perform the DHT
+        f_transform = VM.predict_visibilities(y, grid, k=None, geometry=geom)
+
+    else:
+        y = np.interp(DHT.q, x, f)
+
+        if grid is None:
+            grid = DHT.r * rad_to_arcsec
```
rbooth200, line 916:
```suggestion
            grid = VM.r
```

rbooth200 (dismissed):
I've suggested a few changes that I think should be made.

rbooth200:
I was looking at the inconsistency between what `transform` and `predict_visibilities` do in `VisibilityMapping` and wondering whether I should add an `invert_visiabilities` method VisibilityMapping that is the inverse of `predict_visibilities`. I would keep `transform` too, which does a slightly different thing (i.e. `transform` does a raw DHT while `predict_visibilities`/`invert_visibilities` does additional scaling depending on whether the disc is thick or thin).

frank/utilities.py
jeffjennings: change DHT call to VM in generic_dht

frank/utilities.py
jeffjennings: change a 2nd DHT call to VM in generic_dht

jeffjennings:
I guess I'd remove `transform` and add a bool param `scale_flux` to `predict_visibilities` and `invert_visibilities` that allows you to bypass the scaling for the optically thick case in `_get_mapping_coefficients` e.g. by passing in a `geometry` with all 0s. Then `predict_visibilities` and `invert_visibilities` have the functionality of `transform` while also keeping the option to do the chunking.  I guess `_get_mapping_coefficients` then needs an inverse scale for going from V --> I.

rbooth200:
Ok, so I've added an invert_visibilities method. I've kept the "transform" method because it's used elsewhere in the code base. The difference between the routines is subtle of course, but I think that is why it makes sense to keep them both, and to keep them in the same place.

rbooth200:
@jeffjennings If you're happy with my updated version, feel free to review it yourself and merge. I'm happy to discuss any further changes. One thing to consider is generic_dht the correct name here? The function is doing (slightly) more than just a dht.

jeffjennings:
Can you let me handle generic_dht? We're both editing at same time.

rbooth200:
Yeah, sorry. I'm done.

jeffjennings:
For the inverse transform, I'm getting a correct result when 
```
f_transform = VM.transform(y, grid, direction='backward')
```
but an incorrect result when 
```
f_transform = VM.invert_visibilities(y, grid, geometry=geom)
```
I can't see why when looking at `invert_visibilities` and `_get_mapping_coefficients`. Is it the same for you?

jeffjennings:
> 

I'd like to keep the name; with the default `inc=0`, this is just doing a DHT (after interpolating the input profile onto collocation points, which is a necessary part of taking a DHT of an arbitrary function).

jeffjennings:
import numpy as np 
import matplotlib.pyplot as plt 

from frank.geometry import FixedGeometry, rescale_total_flux
from frank.statistical_models import VisibilityMapping
from frank.utilities import generic_dht

def gauss(x, a, x0, sigma):
    return a * np.exp(-(x - x0)**2 / (2 * sigma**2))
def sigma(fwhm):
    return fwhm / (8 * np.sqrt(np.log(2)))

r = np.linspace(0, 1.0, 1000)

Itrue = gauss(r, 1e10, 0.0, sigma(1.5)) - \
         gauss(r, 3.5e9, 0.3, sigma(0.05))

# load the DSHARP AS 209 observations
as209_dat = np.load('./docs/tutorials/AS209_continuum.npz')
u, v, vis, weights = [as209_dat[k] for k in ['u', 'v', 'V', 'weights']]
sigmas = 1 / np.sqrt(weights)

# deproject, phase-center, rescale the visibilities according to published geometry
geom = FixedGeometry(inc=34.97, PA=85.76, dRA=1.9e-3, dDec=2.5e-3)
ud, vd, visd = geom.apply_correction(u, v, vis) 
visd, weightsd = rescale_total_flux(visd, weights, inc=34.97)

baselines = np.hypot(ud, vd)

grid, ft = generic_dht(r, Itrue, Rmax=2.0, N=1000, direction='forward', grid=None, inc=0.0)
grid2, ft2 = generic_dht(grid, ft, Rmax=2.0, N=1000, direction='backward', grid=None, inc=0.0)

plt.figure()
plt.subplot(121)
plt.plot(r, Itrue, 'k')
plt.plot(grid2, ft2, 'r--')
plt.xlim(0,1)

plt.subplot(122)
plt.plot(grid, ft, 'k')
plt.xlim(-1e6,1e7)
plt.show()

rbooth200:
It was a units error, which should be working now.

jeffjennings:
I see now. Thank you. Is it worth adding `VisibilityMapping.map_visibilities`, `.predict_visibilities` and `.invert_visibilties` to the tests?

rbooth200:
Certainly

jeffjennings:
Do you want me to or you?

rbooth200:
I can do it..

jeffjennings:
Ok thanks. Otherwise I think this is ready to merge. Our changes to `generic_dht` were slightly different but trivially.

rbooth200:
Ok, done. Merge if you are happy

jeffjennings:
Great, thank you for adding the tests for `generic_dht` too.

rbooth200 (approved):
Done

PR #188, opened 2022-11-10
Mock data routines

jeffjennings:
Adds routines to generate mock visibilities and to add Gaussian noise to visibilities

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -893,6 +893,63 @@ def add_vis_noise(vis, weights, seed=None):
 
     return vis_noisy
 
+
+def make_mock_data(r, I, Rmax, u, v, geometry=None, N=500, add_noise=False,
+                   weights=None, seed=None):
+    r"""
+    Generate mock visibilities from a provided brightness profile and (u,v)
+    distribution.
+
+    Parameters
+    ----------
+    r : array, unit = [arcsec]
+        Radial coordinates of I(r)
+    I : array, unit = [Jy / sr]
+        Brightness values at r
+    Rmax : float, unit = [arcsec], default=2.0
+        Maximum radius beyond which I(r) is zero. This should be larger than the
+        disk size
+    u, v : array, unit = :math:`\lambda`
+        u and v coordinates of observations
+    geometry : SourceGeometry object, default=None
+        Source geometry (see frank.geometry.SourceGeometry). If supplied, the
+        visibilities will be deprojected, and their total flux scaled by the
+        inclination.
+    N : integer, default=500
+        Number of terms to use in the Fourier-Bessel series
+    add_noise : bool, default = False
+        Whether to add noise to the mock visibilities
+    weights : array, unit = Jy^-2
+        Visibility weights, of the form :math:`1 / \sigma^2`.
+        If provided, injected noise will be scaled proportional to `\sigma`.
+    seed : int, default = None
+        Number to initialize a pseudorandom number generator for the noise draws
+
+    Returns
+    -------
+    baselines : array, unit = :math:`\lambda`
+        Baseline coordinates of the mock visibilities. These will be equal to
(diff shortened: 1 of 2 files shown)


jeffjennings:
I know that `generic_dht` and `make_mock_data` don't handle the geometrically thick case. And `make_mock_data` takes in `r` and `I` in I(r) as arrays rather than a function. Though I prefer this because it's more generic if slightly less convenient (e.g. if using a profile extracted from an image to generate mock visibilities; you could still just use `generic_dht` and `add_vis_noise` for that, but I think passing in arrays isn't too inconvenient).

frank/utilities.py
```diff
@@ -856,6 +856,100 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
...
+        Weights on the visibilities, of the form :math:`1 / \sigma^2`.
+        Injected noise will be scaled proportional to `\sigma`.
+    seed : int, default = None
+        Number to initialize a pseudorandom number generator for the noise draws
+
+    Returns
+    -------
+    vis_noisy : array, shape = (2, len(V)) if V is complex; else (1, len(V))
+        Visibilities with additive noise
+
```
rbooth200, line 878:
```suggestion
    vis_noisy : array, shape = vis.shape
        Visibilities with added noise

```
The noise shape is 2, len(vis) when vis is complex, but the shape of vis remains the same since you just added the noise to the real/complex parts.

frank/utilities.py
```diff
@@ -856,6 +856,100 @@ def convolve_profile(r, I, disc_i, disc_pa, clean_beam,
...
+    """
+    if seed is not None:
+        np.random.seed(seed)
+
+    dim0 = 1
+    if np.iscomplexobj(vis):
+        dim0 = 2
+
+    noise = np.random.standard_normal([dim0, len(vis)])
```
rbooth200, line 887:
```suggestion
    vis = np.array(vis)
    noise = np.random.standard_normal((dim0,) + vis.shape)
```
This allows us to support arbitrary shapes for vis.

rbooth200 (changes requested):
So I'm mostly happy with this. I've added a couple of small changes that I think make sense. 

There is one other thing that I think we should do, given your preference for having arrays rather than functions as arguments. That is to add one more function that provides the collocation points that are optimal for the DHT. My motivation here is that a bad choice of points will lead to a bad estimate of the DHT. If you only have a poorly sampled array there is not much you can do, but if you have a function you might as well use the optimal points. 

Can you add this function and some notes somewhere along these lines? I don't have time right now, but could do it later today if needed.

jeffjennings:
Ok I added a note to the docstrings along this point, and I can add another function to interpolate onto the collocation points if you prefer, although this is already done in `make_mock_data` because that calls `generic_dht`, which always does the interpolation before taking a transform.

rbooth200:
I don't think we want another interpolation routine, and I know that generic_dht does in interpolation. What I was thinking more was a function that we could use like: 
```
r = get_collocation_points(Rmax, N)
I = f(r)
generic_dht(r, I, ...)
```
Does that make sense?

jeffjennings:
Ah ok I see what you mean. Added in 3c5106a.

rbooth200:
Great, thanks. I added a class method to DHT to compute the collocation points without creating the full object, which is unnecessary.

rbooth200:
Actually, in my mind "forward" should return the radial collocation points since you need those for the I(r) when doing the forward transform. Does that make sense? Or if thats confusing should we use labels for real / fourier space insteafd of transform direction?

jeffjennings:
It makes sense to keep forward/backward I think, it's just early and I was thinking about it 'backward'. Fixed in c2288a3, clarified in the docstring

rbooth200 (approved):

rbooth200:
Feel free to merge if you are happy.

PR #200, opened 2023-03-02
Add non-negative solution to figs; add non-neg. test

jeffjennings:
- Makes `nonnegative` param in .json true by default (as this just calls `setattr(sol, '_nonneg', sol.solve_non_negative())` in `fit.py`)
  * adds check in `fit.py` for whether the fit method is also `Normal`
  
- Adds `sol.solve_non_negative` to quick and full figures      
  *  cleans up some general plot settings, figure syntax

- Adds test for `sol.solve_non_negative`

### DIFF
--- a/frank/fit.py
+++ b/frank/fit.py
@@ -152,6 +152,12 @@ def parse_parameters(*args):
         err = ValueError("method should be 'Normal' or 'LogNormal'")
         raise err
 
+    if model['hyperparameters']['nonnegative'] and \
+       model['hyperparameters']['method'] != 'Normal':
+        err = ValueError("'nonnegative' should only be 'true' if 'method' is "
+                        "'Normal'")
+        raise err
+
     if model['hyperparameters']['method'] == 'LogNormal' and \
        model['hyperparameters']['p0'] is not None and \
        model['hyperparameters']['p0'] > 1e-30:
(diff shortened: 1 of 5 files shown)


jeffjennings:
New versions of quick and full figs look like (from tests):
small_uv_frank_fit_quick
small_uv_frank_fit_full

frank/make_figs.py
```diff
@@ -262,6 +262,11 @@ def make_quick_fig(u, v, vis, weights, sol, bin_widths, dist=None,
 
         plot_brightness_profile(sol.r, sol.I / 1e10, ax1, c='r', label='frank')
 
+        if hasattr(sol, '_nonneg'): 
+            plot_brightness_profile(sol.r, sol._nonneg / 1e10, ax0, ls='--', c='#009933', label='non-neg.')
+            plot_brightness_profile(sol.r, sol._nonneg / 1e10, ax1, ls='--', c='#009933', label='non-neg.')
+
+
```
rbooth200, line 269: I know it's from old code, but it would be better to have not have `nonneg` private (`_nonneg`) because you want people to access it.

rbooth200 (approved):
I've left a suggested change, but you can go with this version if you prefer.

frank/fit.py
```diff
@@ -152,6 +152,12 @@ def parse_parameters(*args):
         err = ValueError("method should be 'Normal' or 'LogNormal'")
         raise err
 
+    if model['hyperparameters']['nonnegative'] and \
+       model['hyperparameters']['method'] != 'Normal':
+        err = ValueError("'nonnegative' should only be 'true' if 'method' is "
+                        "'Normal'")
+        raise err
+
```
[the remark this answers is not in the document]
jeffjennings: Good point, changed to warning in [10613e8](https://github.com/discsim/frank/pull/200/commits/10613e87c47338eadaf25a0ae67e92e461bf3f0b)

frank/make_figs.py
jeffjennings: Good point, thanks! Changed to non-hidden in [22907d7](https://github.com/discsim/frank/pull/200/commits/22907d79ec3128f51103ab76a8d949d632ab2ac9)

jeffjennings:
I applied your suggestions, which made the review stale. Re-approve/merge when you're ready. Thanks!

rbooth200 (approved):

PR #201, opened 2023-03-06
Statistical uncertainty recovery function

jeffjennings:
- Adds a function to `utilities` that calculates statistical uncertainty of a fitted I(r). 

- Adds the `method` key to the `_info` property of `FrankFitter`

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -169,7 +169,7 @@ def get_fit_stat_uncer(fit):
         # see https://en.wikipedia.org/wiki/Log-normal_distribution
         log_variance = np.diag(fit.covariance)
         lin_variance = (np.exp(log_variance) - 1) * \
-            np.exp(2 * np.mean(np.log(fit.I)) + log_variance)
+            np.exp(2 * np.log(fit.I) + log_variance)
         sigma = np.sqrt(lin_variance)
 
     else: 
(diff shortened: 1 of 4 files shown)


frank/utilities.py
```diff
@@ -139,6 +139,46 @@ def jy_convert(x, conversion, bmaj=None, bmin=None):
...
+
+    if fit._info["method"] == "Normal":
+        sigma = np.sqrt(np.diag(fit.covariance))
+
+    elif fit._info["method"] == "LogNormal":
+        # convert stored variance from var(log(brightness)) to var(brightness).
+        # see https://en.wikipedia.org/wiki/Log-normal_distribution
+        log_variance = np.diag(fit.covariance)
+        lin_variance = (np.exp(log_variance) - 1) * \
+            np.exp(2 * np.mean(np.log(fit.I)) + log_variance)
```
rbooth200, line 172:
```suggestion
            np.exp(2 * np.log(fit.I) + log_variance)
```
np.mean here is not correct because it is giving the mean across different radii. The mean provided by wiki would be the posterior mean of log(I) at each radius (assuming that they are independent). 

Note that there is also an argument for neglecting the log_variance term in this line: [see Appendix A.2 here](https://ui.adsabs.harvard.edu/abs/2016A%26A...586A..76J/abstract). I do not have a strong opinion either way here.

rbooth200 (changes requested):

frank/utilities.py
jeffjennings: Sounds good, thanks for catching that and for pointing to the appendix. I've dropped the log_variance term in [67343a7](https://github.com/discsim/frank/pull/201/commits/67343a75d67b750e9362765542ead24ecd432c7c)

rbooth200 (approved):

PR #203, opened 2023-03-13
Raise error on convergence failure

### LINKED ISSUE #202: Throw an error on non-convergence
With `alpha~1` frank can need more than the default `max_iter=2000` to converge. We'd never really worried about this because the best fit is typically hardly changing long before we accept convergence so for practical uses it had been fine to print a warning and rumble on.

I've now come across a case where it is not fine to rumble on, in the work on the debris disc fitter paper with Seba Marino. We're using the posterior probability of the best fit and this is not reliable if we haven't reached convergence.

I'm now preposing that we change the default behaviour to raise an error if the code has not converged in max_iter, while giving users an option to skip it, just printing a warning if convergence hasn't been reached. This would be the 'pythonic' solution. Whatever we decide on, we need a change because currently the code doesn't even print a warning if logging is not enabled.

I'm happy to implement a solution when agreed on.

rbooth200:
This PR has been introduced to fix #202. The options still need adding to fit.py and the parameter file

### DIFF
--- a/frank/radial_fitters.py
+++ b/frank/radial_fitters.py
@@ -810,7 +810,7 @@ class FrankFitter(FourierBesselFitter):
             msg += 'be very slow for alpha close to 1.'
             if self._convegence_failure == 'raise':
                 msg += '\nAlternatively set convergence_failure to'
-                msg += "'warn' or 'ignore' and to continue despite the"
+                msg += "'warn' or 'ignore' to continue despite the"
                 msg += 'failure.'
                 raise RuntimeError(msg)
             elif self._convegence_failure == 'warn':
--- a/frank/radial_fitters.py
+++ b/frank/radial_fitters.py
@@ -678,8 +678,14 @@ class FrankFitter(FourierBesselFitter):
         Specifies the vertical thickness of disc as a function of radius. Both
         R and H should be in arcsec. Assumes a Gaussian vertical structure. 
         Only works with assume_optically_thick=False
-    verbose:
+    verbose: bool
         Whether to print notification messages
+    convergence_failure: string, default = 'raise'
+        Decide what to do when the frank model does not converge within max_iter.
+        Should be one of:
+        'raise'  : raise an error
+        'warn'   : print a warning message and continue
+        'ignore' : Ignore the error.
 
     References
     ----------
@@ -693,7 +699,7 @@ class FrankFitter(FourierBesselFitter):
                  block_size=10 ** 5, alpha=1.05, p_0=None, weights_smooth=1e-4,
                  tol=1e-3, method='Normal', I_scale=1e5, max_iter=2000, check_qbounds=True,
                  store_iteration_diagnostics=False, assume_optically_thick=True,
-                 scale_height=None, verbose=True):
+                 scale_height=None, verbose=True, convergence_failure='raise'):
 
         if method not in {'Normal', 'LogNormal'}:
             raise ValueError('FrankFitter supports following mehods:\n\t'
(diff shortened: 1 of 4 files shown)


frank/radial_fitters.py
```diff
@@ -780,17 +791,36 @@ def _fit(self):
...
                              ' max_iter specified in your parameter file,'
                              ' {}'.format(self._max_iter))
 
+            msg = f'Convergence not met within {self._max_iter} '
+            msg += f'iterations.\nTry increasing max_iter, or '
+            msg += 'try increasing alpha since convergence can '
+            msg += 'be very slow for alpha close to 1.'
+            if self._convegence_failure == 'raise':
+                msg += '\nAlternatively set convergence_failure to'
+                msg += "'warn' or 'ignore' and to continue despite the"
```
jeffjennings, line 813:
```suggestion
                msg += "'warn' or 'ignore' to continue despite the"
```

frank/radial_fitters.py
```diff
@@ -693,7 +699,7 @@ def __init__(self, Rmax, N, geometry, nu=0, block_data=True,
                  block_size=10 ** 5, alpha=1.05, p_0=None, weights_smooth=1e-4,
                  tol=1e-3, method='Normal', I_scale=1e5, max_iter=2000, check_qbounds=True,
                  store_iteration_diagnostics=False, assume_optically_thick=True,
-                 scale_height=None, verbose=True):
+                 scale_height=None, verbose=True, convergence_failure='raise'):
```
jeffjennings, line 702: Better if the default is 'warn', since for most (non-debris) fits the fit is only marginally changing at large iteration, but set it as `raise` in `FrankDebrisFitter`?

frank/radial_fitters.py
```diff
@@ -780,17 +791,36 @@ def _fit(self):
...
+                print()
                 logging.info('    Convergence criterion not met; fit stopped at'
                              ' max_iter specified in your parameter file,'
                              ' {}'.format(self._max_iter))
 
+            msg = f'Convergence not met within {self._max_iter} '
+            msg += f'iterations.\nTry increasing max_iter, or '
+            msg += 'try increasing alpha since convergence can '
+            msg += 'be very slow for alpha close to 1.'
+            if self._convegence_failure == 'raise':
```
jeffjennings, line 811:
```suggestion
            if self._convergence_failure == 'raise':
```

frank/radial_fitters.py
```diff
@@ -780,17 +791,36 @@ def _fit(self):
...
+                msg += '\nAlternatively set convergence_failure to'
+                msg += "'warn' or 'ignore' and to continue despite the"
+                msg += 'failure.'
+                raise RuntimeError(msg)
+            elif self._convegence_failure == 'warn':
```
jeffjennings, line 816:
```suggestion
            elif self._convergence_failure == 'warn':
```

frank/radial_fitters.py
```diff
@@ -780,17 +791,36 @@ def _fit(self):
...
+                if logging.getLogger().isEnabledFor(logging.INFO):
+                    logging.info(msg)
+                else:    
+                    print(msg)
+            elif self._convegence_failure == 'ignore':
```
jeffjennings, line 821:
```suggestion
            elif self._convergence_failure == 'ignore':           
```

jeffjennings (changes requested):
Great! Just a couple typos and a question about whether the default `convergence_failure` should be `warn` for 'standard' frank fits and `raise` for `FrankDebrisFitter`

jeffjennings:
[0f75843](https://github.com/discsim/frank/pull/203/commits/0f758432823a1301d7cd361c9998784e6ee6333b) and [0f75843](https://github.com/discsim/frank/pull/203/commits/0f758432823a1301d7cd361c9998784e6ee6333b) add the param to `fit.py`.

rbooth200:
I prefer raise as default since it's more [pythonic](https://peps.python.org/pep-0020/). Feel free to change it though if you feel strongly. It reminds me that I probably need to add the option to DebrisDiscFitter either way,

jeffjennings:
Sounds good, having it be default raise but default warn in fit.py.

jeffjennings (approved):
Feel free to merge now or after adding the arg to the debris fitters.

rbooth200:
Merging this now because the debris fitters are not in the branch. I'll update it there

PR #205, opened 2023-03-31
Add `I_scale` arg to `.json`s

jeffjennings:
For command-line fits with the `LogNormal` model, the `I_scale` parameter is useful to pass through the `.json` parameter file (for sufficiently faint sources such as some debris disks, the default `I_scale = 1e5` in `FrankFitter` can be too large). PR adds this arg to the `.json`s and the fit pipeline in `fit.py`.

### DIFF
--- a/frank/default_parameters.json
+++ b/frank/default_parameters.json
@@ -39,7 +39,8 @@
     "max_iter"         : 2000,
     "converge_failure" : "warn",
     "nonnegative"      : true,
-    "method"           : "Normal"
+    "method"           : "Normal",
+    "I_scale"          : 1e5
   },
 
   "plotting" : {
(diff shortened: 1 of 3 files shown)


jeffjennings:
Is it worth changing the default `I_scale` in `FrankFitter` and `FrankDebrisFitter` from 1e5 to a much lower value like 1e1 or 1e2? Maybe you've found 1e5 to be more stable or preferable in general relative to a lower value?

rbooth200:
Yeah convergence can be slow with a bad choice of I_scale, but a too large value is really not good. I'm not sure what is optimal for a large range of datasets though.

rbooth200 (approved):

PR #207, opened 2023-07-03
save_fit: save correct profile uncertainty

### LINKED ISSUE #206: error I(r) in save_fit wrong when using method lognormal
The error on I(r) is calculated as np.diag(sol.covariance)**.5]).T when saving the radial profile in save_fit() in the main and debris_update branches. This would be wrong when using the lognormal method.

jeffjennings:
Ensures that `io.save_fit` stores the correct brightness profile uncertainty regardless of whether the fit method is `Normal` or `LogNormal`.

Fixes #206

### DIFF
--- a/frank/io.py
+++ b/frank/io.py
@@ -25,6 +25,8 @@ import numpy as np
 import pickle
 import logging
 
+from frank.utilities import get_fit_stat_uncer
+
 def load_uvtable(data_file):
     r"""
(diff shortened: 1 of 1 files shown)


rbooth200 (approved):

rbooth200:
I merged this. We should merge the debris branch too.

PR #209, opened 2023-07-06
v1.2.2

jeffjennings:
There have been a decent amount of changes (~100 commits) since v.1.2.1, including some code refactoring, additional debris fitter integration and one bugfix. I think it would be good to bump the version so the pip-installable frank has these changes.


rbooth200 (approved):

rbooth200:
This is a good idea, merge when ready.

PR #211, opened 2023-10-26
Bug in sweep_profile args

jeffjennings:
Calling `utilities.sweep_profile` with `project=False` and `phase_shift=True` raises an error because `dra` and `ddec` are defined internally only if `phase_shift = True`. PR fixes this.


rbooth200 (approved):

rbooth200:
Looks good to me

PR #212, opened 2023-10-26
Update atleast_1d syntax for new numpy versions

### LINKED ISSUE #210: wrong syntax for np.atleast_1d ?
line 426 of utilities.py

`np.atleast_1d(wle, dtype='f8')` raises TypeError

should maybe be `np.atleast_1d(wle).astype('f8')` ?

jeffjennings:
`atleast_1d` call signature has changed in recent numpy versions (confirmed on v1.21.5; current version is 1.26.1). PR updates call signature to not rasie error in `utilities.normalize_uv`

Fixes #210

### DIFF
--- a/frank/utilities.py
+++ b/frank/utilities.py
@@ -423,7 +423,7 @@ def normalize_uv(u, v, wle):
     logging.info('  Normalizing u and v coordinates by provided'
                  ' observing wavelength of {} m'.format(wle))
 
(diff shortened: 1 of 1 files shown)


rbooth200 (approved):
This looks fine. 

I guess before this was only working by chance - numpy doesn't seem to have changed their API, just their implementation.

I don't see why the astype('f8') is really needed, but I don't see any harm in it either.

PR #213, opened 2023-11-01
Expand `make_mock_data` projection options

jeffjennings:
Adds option in `make_mock_data` to reproject rather than deprojected the generated mock data. Handles the expected behavior of the function from its previous implementation (when only `geometry` was passed in order to deproject). 

This option is useful to quickly generate reprojected mock visibilities from a deprojected frank brightness profile in order to compare to observations or to external software like [MPoL](https://github.com/MPoL-dev/MPoL) that works with projected data and performs its own implementation of deprojection.


docs/tutorials/mock_data.ipynb
```diff
@@ -230,11 +230,13 @@
    "source": [
     "# sample the mock profile at the deprojected data baselines, \n",
     "# adding Gaussian noise according to the data weights\n",
-    "baselines, V_noisy = make_mock_data(r, I, Rmax=2.0, u=u, v=v, geometry=geom, \n",
```
jeffjennings, line 233: Location of changes in mock_data.ipynb tutorial

rbooth200 (approved):
This looks fine. We have `fit.predict` that can be used when you have a fit to real data, but this is doing something slightly different.

jeffjennings:
Great, thanks!

PR #215, opened 2023-11-02
Utils tests

### LINKED ISSUE #214: Broader test coverage
**Describe the bug -**
#211 introduced a bug in `utilities.sweep_profile` due to a typo ('geom.dra' instead of 'geom.dRA'); this was my fault, but in general it would be more consistent to catch simple errors with better test coverage. `utilities` has the worst test coverage.

**Additional context -**
I suggest adding tests for the routines in `utilities` that aren't covered by existing tests. I'll create a PR that does this.

jeffjennings:
Adds tests for the routines in `utilities.py` that aren't covered by existing tests. These tests check that the routines simply run and that they give the expected numerical results.

Closes #214

### DIFF
--- a/frank/tests.py
+++ b/frank/tests.py
@@ -618,6 +618,22 @@ def test_add_vis_noise():
     np.testing.assert_almost_equal(result, [-0.0992665+2.74683048j])
 
 
+def test_make_mock_data():
+    """Check utilities.add_vis_noise"""
+    AS209, AS209_geometry = load_AS209(uv_cut=1e6)
+    u, v, vis, weights = [AS209[k][::100] for k in ['u', 'v', 'V', 'weights']]
+    
+    # generate a sol from a standard frank fit
+    FF = FrankFitter(1.6, 20, AS209_geometry, alpha=1.3, weights_smooth=1e-2)
+    sol = FF.fit(u, v, vis, weights)   
+
+    # call with minimal inputs
+    result = utilities.make_mock_data(sol.r, sol.I, 3.0, u, v)
+    expected = [ 0.06699455,  0.18302498,  0.27758414,  0.04789997, -0.00240399,
+        0.06339718,  0.00358722,  0.28862088,  0.07058801,  0.06617371]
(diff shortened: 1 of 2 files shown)


jeffjennings:
The other program that has low test coverage is `filter.py`; we should probably add tests for those missing routines (three of the functions in `CriticalFilter`). I can do that if you don't feel like it (or unless you expect those functions to change).

rbooth200:
I can add regression tests for two of the functions in filter.py. The function `update_power_spectrum_multiple` isn't currently used anywhere so I'd leave it untested. Can you check the maths for the other two? The covariance function should be the same as the Hessian in the frank debris paper and the prior is log `P(p|beta)` in the frank paper.

jeffjennings:
> I can add regression tests for two of the functions in filter.py. The function `update_power_spectrum_multiple` isn't currently used anywhere so I'd leave it untested. Can you check the maths for the other two? The covariance function should be the same as the Hessian in the frank debris paper and the prior is log `P(p|beta)` in the frank paper.

Great, changes to `filter.py` should probably be a separate PR. 

Re: math: for `log_prior`, in the docstring should `- np.sum(p0/pi + (alpha-1)*np.log(p0/pi))` be `- np.sum(p0/pi + (alpha-1)***np.log(pi)**`  -- (eqns.37-40 in the paper)? If it's correct as it is, the docstring and code are consistent.

For `covariance_MAP`, (eqn.40 in the debris paper), I'm confused where the 2 comes from in `- 0.5 * np.outer(1 / p, 1 / p) * (**2** * mqq + Dqq) * Dqq`, but otherwise it looks right.

rbooth200:
I checked that term - it's the first line in eqn 30, so I think everything is fine. I've added a test that checks both of the missing functions in `filter.py` and some of the other likelihood functions. We're now up to 86% coverage with ~half the missing lines coming from stats models and make_figs

rbooth200 (approved):
Approving this. If you're happy with the filter tests, feel free to merge.

jeffjennings:
Great, thank you

PR #219, opened 2024-08-21
solve_non_negative bug; increment frank version

jeffjennings:
This PR was at first meant to step frank from v1.2.2 to 1.2.3 after someone was trying to use the pip version of frank and encountered a bug that we fixed after that release. I figured I'd also bump the python versions to drop support for 3.7 and add support for 3.12.

However, a separate problem arose - the current [test](https://github.com/discsim/frank/blob/9de2461579c95ab3b6c484be4819f4681a735967/frank/tests.py#L317) for [`solve_non_negative`](https://github.com/discsim/frank/blob/9de2461579c95ab3b6c484be4819f4681a735967/frank/statistical_models.py#L838) passes on python 3.8 but not 3.9+, returning an array of `0`.  The underlying scipy [implementation](https://github.com/scipy/scipy/blob/v1.14.0/scipy/optimize/_nnls.py#L8-L95) was updated ~6 months ago. @rbooth200, do you have an idea for a quick fix?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: 3.7
+          python-version: 3.12
 
       - name: Display Python version
         run: python -c "import sys; print(sys.version)"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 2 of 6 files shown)


rbooth200:
I think we should fix #222 before merging this one though.

rbooth200 (dismissed):

jeffjennings:
Ok I updated the changelog with our changes in the last couple days. I'll wait to merge this and release a new version until #224 and #225 are merged.

rbooth200 (approved):

PR #221, opened 2024-08-21
`GaussianModel`: update failed fit error message

### LINKED ISSUE #218: Warning about bad power spectrum
I was getting the following warning for certain choices of parameters: "Bad value in power spectrum. The power spectrum must be positive and not contain any NaN values. This is likely due to your UVtable (incorrect units or weights), or the deprojection being applied (incorrect geometry and/or phase center). Else you may want to increase `rout` by 10-20% or `n` so that it is large, >~300." In my case, I found that _decreasing_ rather than increasing r_out fixed things, so it may be helpful to users to make the warning more general.

jeffjennings:
Generalize message so as to be more helpful.

closes #218 

Tests will fail until #222 is addressed.

### DIFF
--- a/frank/statistical_models.py
+++ b/frank/statistical_models.py
@@ -674,8 +674,8 @@ class GaussianModel:
                                  " your UVtable (incorrect units or weights), "
                                  " or the deprojection being applied (incorrect"
                                  " geometry and/or phase center). Else you may"
-                                 " want to increase `rout` by 10-20% or `n` so"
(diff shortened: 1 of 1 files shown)


rbooth200:
I'm happy to merge this. It's not the cause of the failures, certainly.

rbooth200 (approved):

jeffjennings:
Ok great, thanks!

PR #223, opened 2024-08-22
Fix scipy nnls issue

### LINKED ISSUE #222: tests failure due to `solve_non_negative`
The current [test](https://github.com/discsim/frank/blob/9de2461579c95ab3b6c484be4819f4681a735967/frank/tests.py#L317) for [solve_non_negative](https://github.com/discsim/frank/blob/9de2461579c95ab3b6c484be4819f4681a735967/frank/statistical_models.py#L838) passes on python 3.8 but not 3.9+, returning an array of 0. The underlying scipy [implementation](https://github.com/scipy/scipy/blob/v1.14.0/scipy/optimize/_nnls.py#L8-L95) was updated ~6 months ago. @rbooth200, do you have an idea for a quick fix?

rbooth200:
This PR tries to fix #222 by excluding the broken scipy versions.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -39,7 +39,7 @@ python_requires = >=3.7
 install_requires =
     numpy>=1.12
     matplotlib>=3.1.0
-    scipy>=1.2.0
+    scipy>=1.2.0, !=1.12.*, !=1.13.*,!=1.14.*
 
 # additional files to include in install
 [options.package_data]


rbooth200:
This PR definitely works for now. We should probably check back that everything is still ok when the next scipy version drops.

jeffjennings:
Great, thank you!

PR #230, opened 2024-10-31
Fixed typo in geometry printing

rbooth200:
I spotted a typo when doing `print(geometry)`. This PR makes it so that the output can be simply copied to create a FixedGeometry object, which is useful, e.g., when fitting the geometry once.


jeffjennings (approved):

jeffjennings:
Great!

